Issue #1, opened 2026-01-04
How does this handle interactive questions and prompts from Claude?


ColbySerpa:
In Claude Code CLI, Claude often asks the user questions that depend on using up down left right arrow keys... How is that handled without CLI? Is that not addressed yet?

ColbySerpa:
> > Is that available in this SDK as well?
> 
> Yeah you would do something like this: 
> ```
> qs.Q(1).Text("custom response")
> ```

It usually offers 3 options, the first two are selectable and the 3rd option gives you an option for a free-written custom response.

Although, I'm worried -- I suspect no SDKs can stream the real-time token counter etc since they work via JSON. It would be creating so many events per second.

Roasbeef:
Fixed by #5

PR #5, opened 2026-01-05
multi: add native support for Q&A w/ claude

Roasbeef:
This PR adds support for handling Claude's AskUserQuestion tool programmatically, addressing issue #1. When Claude needs to ask clarifying questions during a conversation, the SDK now provides two ways to handle them: a callback API for automatic handling, and a QuestionMessage type that appears in the Query() message stream.

The callback API uses WithAskUserQuestionHandler to register a function that receives questions and returns answers. Questions are handled automatically without appearing in the message stream. For more control, callers can omit the handler and receive QuestionMessage instances in Query(), responding inline with the Respond() method.

QuestionSet provides fluent helper methods for constructing answers. Instead of manually building {"q_0": "value"} maps, callers can use qs.Answer(0, "Yes"), qs.Q(0).Select("label"), or qs.Q(0).SelectIndex(0). These helpers reduce boilerplate and prevent key format errors.

The implementation includes comprehensive unit tests for all helper methods and integration tests that verify both the callback and QuestionMessage flows work with the real CLI. Integration tests use a system prompt encouraging Claude to ask questions, though actual behavior depends on the model.

This PR also adds CLAUDE.md with project instructions for AI agents, documenting the distinction between unit tests (go test ./...) and integration tests (make test-integration) which require an API token.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 go build ./...                          # Build
 go test ./...                           # Unit tests
 go test -v -run TestName ./...          # Single test
+make lint                               # Run linter (must pass)
 make test-integration                   # Integration tests (needs API token)
 make test-race                          # Unit tests with race detector
 ```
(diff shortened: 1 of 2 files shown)


Roasbeef:
>  is there handling for displaying when a sub-agent is spun up and checking its activity? @Roasbeef Usually it's ctrl+o to see the details I think. 🤔

Yep this is already built in, made a PR to add a better guide here: https://github.com/Roasbeef/claude-agent-sdk-go/pull/6

PR #47, opened 2026-04-26
transport: extract Transport interface; add EndInput/IsReady

Roasbeef:
Closes #16.

## Summary
Extracts a `Transport` interface so callers can swap the subprocess for in-memory or networked transports without touching `Protocol` or `Client`. `SubprocessTransport` keeps its current semantics and now implements the new interface, picking up `EndInput` (close stdin without tearing down the read side) and `IsReady`.

## Changes
- `transport.go` — new `Transport` interface (`Connect`, `Write`, `ReadMessages`, `EndInput`, `Close`, `IsReady`); `SubprocessTransport` implements it; compile-time assertion via `var _ Transport = (*SubprocessTransport)(nil)`.
- `client.go` — `Client.transport` field is now `Transport`. `Connect` uses `Options.Transport` if set, otherwise builds the default subprocess. Stderr wiring stays subprocess-only (in-memory transports don't have stderr).
- `options.go` — `Options.Transport Transport` field (`json:"-"`) and `WithTransport` functional option.
- `protocol.go` — `Protocol.transport` typed as `Transport`; protocol only ever calls `Write`, fully covered by the interface.
- `transport_test.go` — adds a `mockTransport` that satisfies `Transport`, plus `EndInput` coverage (basic, idempotent, plays nicely with `Close`).

## Notes
- `EndInput` and `Write` share `t.mu`, so they serialize. Calling `Write` after `EndInput` is a contract violation per the interface doc — no defensive guard added; behavior is loud nil-deref by design.
- No behavior change for the default subprocess path. Existing tests continue to construct `SubprocessTransport` directly and exercise the same code paths.

## Test plan
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test ./...`
- [ ] CI green on the PR

### DIFF
--- a/client.go
+++ b/client.go
@@ -16,7 +16,7 @@ import (
 // iter.Seq for streaming message iteration.
 type Client struct {
 	options   Options
-	transport *SubprocessTransport
+	transport Transport
 	protocol  *Protocol
 	skills    []Skill
 	mu        sync.Mutex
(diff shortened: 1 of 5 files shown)


Roasbeef:
VERDICT: REQUEST_CHANGES

The interface and subprocess behavior are on target, but the mock transport path is not actually exercised and CI lint is red.

The Go `Transport` surface matches the TS abstraction in `sdk.d.ts` L5306-L5333 after adapting to Go message types/context, and the core subprocess changes look correct: `EndInput` serializes on `t.mu`, nils stdin, does not flip `closed`, `Close` still guards nil stdin, and `IsReady` delegates to `IsAlive`.

Please add the missing mock transport coverage from the PR brief: at minimum a direct `mockTransport.ReadMessages` round-trip test that uses `newMockTransport`, pushes messages through `incoming`, drives the iterator, and proves repeated `Close` does not panic. A `Client` + `WithTransport(mock)` path would also cover the injection branch; if that is intentionally carved out, document it in the PR body.

CI is currently failing lint on the same gap: `transport_test.go:74:6: func newMockTransport is unused (unused)`.

Overall: the production refactor is close, but the new interface hook needs exercised mock coverage and a green lint gate before merge.

Roasbeef:
Addressed in `cd5c9a2`:

- `TestMockTransportReadMessagesRoundTrip` — now exercises `newMockTransport`, drives `ReadMessages` end-to-end, asserts ordered yields, then proves repeat `Close` is a no-op and `Write` after `Close` returns `ErrTransportClosed`. This also clears the staticcheck `unused` finding on `newMockTransport`.
- `TestWithTransportOptionPlumbed` — verifies `WithTransport` stores the injected transport on `Options` so `Client.Connect`'s injection branch picks it up.
- PR body updated with the explicit carve-out: full `Client.Connect` + `WithTransport(mock)` end-to-end coverage requires fixturing the control-channel `initialize` handshake against the mock and is deferred to a follow-up.

PR #48, opened 2026-04-26
stream: await SDK control responses (PR 17 of v0.2.119 catchup)

Roasbeef:
## What

Rewires the four `Stream` runtime control methods — `Interrupt`,
`SetPermissionMode`, `SetModel`, `SetMaxThinkingTokens` — to await the matching
`SDKControlResponse` instead of fire-and-forget, matching the TS SDK contract
(`sdk.d.ts` L1960-2001, `sdk.mjs` `await this.request(...)`).

## Why

Pre-PR, each method built an `SDKControlRequest` and called
`transport.Write(ctx, req)` without registering a pending-response slot. That
diverged from upstream in three ways:

- A caller could not tell whether the CLI accepted the change.
- An `error` reply (e.g. invalid permission mode) was silently dropped.
- The pending-request slot in `protocol.pendingReqs` was never registered, so
  even if a response arrived it would be matched against nothing.

This PR fixes all three.

## How

Added a single `sendSDKControlRequest` helper on `*Stream` that:

1. Registers the pending-response slot **before** writing, to avoid a race
   where a fast CLI reply lands before the waiter is in the map.
2. Writes the request via the transport.
3. Selects on `ctx.Done()` and the response channel.
4. Surfaces `error`-subtype responses as a wrapped Go error.

The four methods now delegate to that helper. Helper returns
`*SDKControlResponse` even though PR 17's four callers ignore it — PRs 18–20
will need the parsed body for introspection methods, so no churn later.

### Out of scope (deferred)

`Protocol.waitForSDKResponse` registers the channel **after** the write, which
has the same race the helper here avoids by reversing the order. Not touched
in this PR; flagging for a follow-up.

## Tests

New `stream_control_test.go`. Hand-rolled `*Protocol` + mock transport so the
tests drive the real `Stream` methods end-to-end without spawning a subprocess.

| Test | What it asserts |
|------|-----------------|
| `TestStreamInterruptSendsControlRequest` | Wire shape: `{type:control_request, request:{subtype:interrupt}}`, no spurious mode/model/max_thinking_tokens fields |
| `TestStreamSetPermissionModeSendsModeField` | `subtype=set_permission_mode`, `mode=acceptEdits` |
| `TestStreamSetModelSendsModelField` | `subtype=set_model`, `model=claude-sonnet-4-5-20250929` |
| `TestStreamSetMaxThinkingTokensRoundTrip` | Three subtests: `nil` → field omitted; `*int(0)` → present as 0; `*int(4096)` → present as 4096 |
| `TestStreamControlRequestSurfacesError` | CLI returns `{subtype:error, error:"invalid mode"}` → method returns wrapped error |
| `TestStreamControlRequestRespectsContextCancel` | Cancelling `ctx` returns `context.Canceled` AND deletes the pending slot |

`go build ./...`, `go vet ./...`, `gofmt -l .` clean. `go test -race ./...` green.

## Catchup context

Phase 6, second PR of the v0.2.119 catchup (`@anthropic-ai/claude-agent-sdk@0.2.119`).
This unblocks PRs 18–20 by establishing the await-helper pattern the rest of the
control surface (`SupportedCommands`, `RewindFiles`, `AccountInfo`, etc.) will
adopt.

### DIFF
--- a/client.go
+++ b/client.go
@@ -747,16 +747,12 @@ func (s *Stream) handleSends() {
 }
 
 // Interrupt sends an interrupt signal to stop the current generation.
+// It blocks until the CLI acknowledges the request or returns an error.
 func (s *Stream) Interrupt(ctx context.Context) error {
-	// Send interrupt control message in SDK format.
-	req := SDKControlRequest{
-		Type:      "control_request",
-		RequestID: s.client.protocol.nextRequestID(),
-		Request: SDKControlRequestBody{
-			Subtype: "interrupt",
-		},
-	}
-	return s.client.transport.Write(ctx, req)
+	_, err := s.sendSDKControlRequest(ctx, SDKControlRequestBody{
+		Subtype: "interrupt",
+	})
+	return err
 }
 
 // RewindFiles restores files to a checkpoint at the specified user message.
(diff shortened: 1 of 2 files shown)


Roasbeef:
VERDICT: APPROVE

Self-review (gh token is PR author — posting as comment per self-approval playbook).

- Helper logic correct: register-before-write avoids the documented race in `Protocol.waitForSDKResponse`. Cleanup on write-error and ctx-cancel; success path relies on `handleSDKControlResponse`'s `LoadAndDelete`.
- Four methods rewritten exactly per BRIEF; godoc updated to flag blocking semantics.
- `RewindFiles` intentionally left fire-and-forget (PR 18–20 scope per brief).
- Tests cover all six BRIEF cases including the three `*int` variants for `max_thinking_tokens` and the pending-slot leak check on ctx cancel.
- `go build`, `go vet`, `gofmt -l`, `go test -race ./...` all clean. CI green across linux/darwin/windows + lint + test.

Squash-merging.

PR #50, opened 2026-04-26
stream: MCP runtime control (PR 19 of v0.2.119 catchup)

Roasbeef:
PR 19 of the v0.2.119 catchup. Adds three MCP runtime-control methods to `Stream`, matching the upstream surface in `sdk.d.ts` L2105-2141.

## Methods

| Method | Subtype | Payload |
|---|---|---|
| `ReconnectMcpServer(ctx, serverName)` | `mcp_reconnect` | `serverName` |
| `ToggleMcpServer(ctx, serverName, enabled)` | `mcp_toggle` | `serverName`, `enabled` |
| `SetMcpServers(ctx, servers) (*McpSetServersResult, error)` | `mcp_set_servers` | `servers` |

All three route through `sendSDKControlRequest` (PR 17). `Enabled` is `*bool` so an explicit `false` serializes; `Servers` is `*map[...]` so the field is omitted on unrelated subtypes but a nil/empty input still round-trips as `"servers":{}` to match the TS "remove all dynamic servers" semantics.

## Wire field divergence

`SDKControlRequestBody.ServerName` (snake_case `server_name`) is preserved for `mcp_message`. The new `MCPServerName` field (camelCase `serverName`) is added separately for the MCP control subtypes that use the camelCase form. Same Go-level concept on the wire as two different keys; documented inline.

## Tests

`client_mcp_runtime_test.go` covers, for each method:
- outbound wire shape (subtype + field names, including `enabled:false`)
- response parsing (populated and empty arrays/maps)
- error propagation through `sendSDKControlRequest`

## Out of scope

- In-process SDK MCP server hot-swap (`Options.SDKMcpServers`). TS `setMcpServers` rebuilds the client-side SDK MCP registry; Go's registry is initialized once and never mutated. Adding hot-swap means extending `MCPServerConfig` with an SDK-instance variant and exposing a `Protocol` mutator — separate PR.
- Other Stream methods (`ReadFile`, `ReloadPlugins`, `ApplyFlagSettings`, `StopTask`, `Close` semantics) — PR 20.
- Session-management free functions — PR 21.
- Other MCP subtypes (`mcp_authenticate`, `mcp_clear_auth`, `mcp_oauth_callback_url`) — out of v0.2.119 scope.

### DIFF
--- a/client.go
+++ b/client.go
@@ -938,6 +938,62 @@ func (s *Stream) GetContextUsage(
 	return &out, nil
 }
 
+// ReconnectMcpServer asks the CLI to reconnect the named MCP server.
+// Returns an error if the server is unknown or reconnection fails.
+func (s *Stream) ReconnectMcpServer(ctx context.Context, serverName string) error {
+	_, err := s.sendSDKControlRequest(ctx, SDKControlRequestBody{
+		Subtype:       "mcp_reconnect",
+		MCPServerName: serverName,
+	})
+	return err
+}
+
+// ToggleMcpServer enables or disables the named MCP server.
+// Returns an error if the server is unknown or the toggle fails.
+func (s *Stream) ToggleMcpServer(
+	ctx context.Context,
+	serverName string,
+	enabled bool,
+) error {
+	_, err := s.sendSDKControlRequest(ctx, SDKControlRequestBody{
+		Subtype:       "mcp_toggle",
+		MCPServerName: serverName,
+		Enabled:       &enabled,
+	})
+	return err
+}
+
+// SetMcpServers replaces the dynamically-added MCP servers for this session
+// with the supplied set. Servers omitted from the new map are disconnected;
+// new entries are connected.
+//
+// Process-based servers only: stdio, sse, http. In-process SDK MCP servers
+// (Options.SDKMcpServers) are static for the lifetime of the client.
+func (s *Stream) SetMcpServers(
+	ctx context.Context,
+	servers map[string]MCPServerConfig,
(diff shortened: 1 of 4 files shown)


Roasbeef:
VERDICT: REQUEST_CHANGES

Fix the formatting failure before merge; the implementation otherwise matches the PR 19 runtime-control surface.

messages.go:264: lint is failing on gofmt for the SDKControlRequestBody field block. Run gofmt on messages.go so the new MCPServerName/Enabled/Servers fields align with the surrounding struct formatting. CI currently reports: messages.go:264:1: File is not properly formatted (gofmt).

Checked against sdk.d.ts L1009-L1025 and L2105-L2141: McpSetServersResult tags, method signatures, subtype strings, serverName vs server_name split, enabled:false handling, nil/empty servers -> {}, and JSON-tag response parsing all look correct. The required unit coverage is present, including enabled:false, nil and empty server maps, empty result collections, and control-error propagation.

Overall: Requesting changes only for the gofmt/lint failure.

Roasbeef:
Pushed 4c07f6e: dropped the standalone block comment that was splitting the SDKControlRequestBody gofmt alignment group. Moved the wire-protocol divergence note into trailing per-field comments (server_name vs serverName, *bool semantics, *map round-trip). No code changes — just struct comment placement to satisfy gofmt.

Roasbeef:
VERDICT: APPROVE — gofmt fix applied (comment placement only), CI green across lint/build/test. MCP runtime-control surface verified against the brief. Merging.

PR #64, opened 2026-06-01
options: add typed APIProvider with `gateway` variant

Roasbeef:
TS SDK v0.3.150 added `'gateway'` as a 7th value of `AccountInfo.apiProvider` ("the CLI is authenticated against an enterprise gateway"). The Go SDK already carried this as a bare string, so the new value round-trips today.

This PR makes the variant set explicit:
- Add an `APIProvider` string alias and constants for the 7 values (`APIProviderFirstParty` ... `APIProviderGateway`).
- Retype `AccountInfo.APIProvider` from `string` to `APIProvider`.

Behavior unchanged; existing JSON shape unchanged. Tests added for the new `gateway` variant round-trip.

Part of the v0.3.150 catchup (PR 1 of 34).

### DIFF
--- a/client_introspection_test.go
+++ b/client_introspection_test.go
@@ -35,7 +35,7 @@ func canonicalInitResponse() *SDKControlInitializeResponse {
 			SubscriptionType: "pro",
 			TokenSource:      "oauth",
 			APIKeySource:     "user",
-			APIProvider:      "firstParty",
+			APIProvider:      APIProviderFirstParty,
 		},
 		FastModeState: "off",
 	}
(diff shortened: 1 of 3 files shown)


Roasbeef:
## Review (daedalus orchestrator, self-approval-as-comment)

VERDICT: APPROVE

**Scope match:** Diff matches `pr-01-api-provider-gateway/BRIEF.md` exactly — adds typed `APIProvider` alias + 7 constants on `query_types.go`, retypes `AccountInfo.APIProvider`, updates call sites to use constants, adds round-trip tests covering both new `gateway` variant directions.

**Correctness:**
- Constants cover all 7 TS variants verbatim: `firstParty`, `bedrock`, `vertex`, `foundry`, `anthropicAws`, `mantle`, `gateway`. Spot-checked against `scratch/ts-sdk-0.3.150/package/sdk.d.ts`.
- `string`-alias keeps the JSON shape stable; existing producers/consumers compile through. Confirmed by the unmarshal round-trip on raw `{"apiProvider":"gateway","apiKeySource":"oauth"}`.
- Integration test smoke-checks the returned provider against the known constant set without hard-failing on empty (OAuth-token-only paths can return empty AccountInfo).

**Style:** doc comment on `APIProvider` follows Go conventions; gofmt clean; go vet clean; `go test ./...` green locally.

**Minor (non-blocking):** `TestAccountInfoAPIProviderGatewayUnmarshal` largely overlaps with the gateway row in `TestAccountInfoAPIProviderJSON` — only the `apiKeySource` co-field check is unique. Could fold into the table in a follow-up sweep; not worth a respin.

Merging on CI green per the self-approval playbook.

PR #66, opened 2026-06-01
options: add MCPServerConfig.Timeout for per-server tool-call timeout

Roasbeef:
## What

Adds `Timeout *int` (milliseconds) on `MCPServerConfig`. Overrides `MCP_TOOL_TIMEOUT` for that one server. Hard wall-clock per call - progress notifications do not extend it. Server-side floor is 1000ms.

## Wire paths

Two places encode `MCPServerConfig`, and both now carry `timeout`:

1. **`--mcp-config`** (`SubprocessTransport.Connect`) - the JSON config blob written to the CLI's `--mcp-config` flag.
2. **Control-init body** (`Protocol.Initialize` -> `SDKControlRequestBody.MCPServers`) - previously missing entirely. Added so the SDK matches the established pattern of duplicating session-shaping options through both the CLI flag and the init body (same shape as `SystemPrompt`, etc.).

`omitempty` semantics: nil timeout is dropped from the wire; explicit `0` is preserved.

## Tests

- `TestMCPServerConfigTimeoutJSON` - JSON round-trip on the struct itself (nil omitted, zero preserved, value round-trips).
- `TestSubprocessTransportMCPConfigEncoding/stdio with timeout` - subprocess `--mcp-config` path.
- `TestProtocolInitializeOptions/mcp server timeout wired through` - control-init body path.

## Followups

`TestIntegrationMCPTimeout` is a `t.Skip()` placeholder. Driving it live needs a slow stdio MCP fixture so the CLI can actually time out a tool call; opening that as separate work.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -602,6 +602,13 @@ func TestIntegrationSDKMCPServer(t *testing.T) {
 		"expected response to contain the sum 11 (7 + 4)")
 }
 
+func TestIntegrationMCPTimeout(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	t.Skip("not directly triggerable from CLI without slow stdio MCP fixture: in-process SDK MCP servers are registered separately from MCPServerConfig timeout")
+}
+
 // TestIntegrationStopHookBlock tests that Stop hooks can block session exit
 // and reinject a new prompt using the Decision/Reason/SystemMessage fields.
 //
--- a/mcp_test.go
+++ b/mcp_test.go
@@ -46,6 +46,60 @@ func TestMCPServerConfigEnv(t *testing.T) {
 	assert.Equal(t, "true", config.Env["DEBUG"])
 }
 
+func TestMCPServerConfigTimeoutJSON(t *testing.T) {
+	ptr := func(v int) *int { return &v }
+
+	t.Run("marshal includes timeout", func(t *testing.T) {
+		data, err := json.Marshal(MCPServerConfig{
+			Type:    "stdio",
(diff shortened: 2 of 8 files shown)


Roasbeef:
VERDICT: APPROVE

Diff matches brief end-to-end:

- `options.go` — `Timeout *int` added to `MCPServerConfig` with the mirrored TS docstring (ms, overrides `MCP_TOOL_TIMEOUT`, wall-clock, 1000ms server floor). `*int` keeps unset vs explicit-zero distinguishable. ✅
- `messages.go` — `SDKControlRequestBody.MCPServers map[string]MCPServerConfig` filled in the previously-missing init-body slot. ✅
- `protocol.go` — `p.options.MCPServers` wired into the initialize control request. ✅
- `transport.go` — timeout serialized into the `--mcp-config` JSON blob. ✅
- Tests — `TestMCPServerConfigTimeoutJSON` correctly probes the `*int + omitempty` semantics (nil dropped, explicit `0` preserved); `TestProtocolInitializeOptions/mcp server timeout wired through` covers the control-init path; `TestSubprocessTransportMCPConfigEncoding/stdio with timeout` covers the subprocess path. ✅
- Integration — `t.Skip` with clear reason (in-process SDK MCP servers register separately from `MCPServerConfig` timeout), matches BRIEF §integration fallback. ✅

CI green across Lint / Test / Build (linux,darwin,windows × amd64,arm64). Integration skipped as designed.

(Posting as comment rather than formal APPROVE because the gh token is the PR author; orchestrator squash-merging per the self-approval-blocker playbook.)

PR #67, opened 2026-06-01
mcp: add per-server alwaysLoad + sdk-mcp instructions

Roasbeef:
In this PR, we wire up two related MCP-startup-time knobs from TS SDK
v0.3.150 (Phase B, PR-07 in the catchup plan). Both fields ship together
because they're MCP-startup-time concerns that touch the same control-init
body shape, and the brief at `memory/catchup-v0.3.150/pr-07-mcp-always-load/BRIEF.md`
intentionally bundles them.

## `MCPServerConfig.AlwaysLoad`

A new `*bool` field on `MCPServerConfig` (`options.go`) that mirrors TS SDK
`McpStdioServerConfig.alwaysLoad`, `McpHttpServerConfig.alwaysLoad`,
`McpSSEServerConfig.alwaysLoad`, and `McpClaudeAIProxyServerConfig.alwaysLoad`.
When true, every tool from this server is included in the prompt instead of
deferred behind tool search, equivalent to setting `defer_loading: false` on
the API. Side effect: setting it blocks startup until the server is connected
(capped at the standard 5s connect timeout) even though MCP startup is
otherwise non-blocking, since the tools must be present when the turn-1 prompt
is built.

`*bool` (not `bool`) so we can distinguish unset (default-defer) from explicit
`false` (reaffirming the default). Same rationale as `Timeout *int` from PR-06.
`omitempty` keeps the existing JSON shape unchanged for configs that don't set
it.

In `transport.go`, we encode `alwaysLoad` into the `--mcp-config` payload
alongside `timeout`, before the type-specific fields, so it applies uniformly
across stdio/http/sse/socket. Same flat-collapse pattern PR-06 used.

## `McpServerOptions.Instructions`

A new `string` field on `McpServerOptions` (`mcp.go`) that mirrors TS SDK
`SdkMcpServerConfig.instructions`. Used when proxying a real MCP server
through the SDK transport so the underlying server's `getInstructions()`
block isn't dropped on the way through.

`CreateMcpServer` stashes it on a new `McpServer.instructions` private field;
we expose it via a new `Instructions()` getter. In `protocol.go`'s
`handleSDKMCPMessage` initialize case, we extract the `result` map out of the
JSONRPC envelope and conditionally set `result["instructions"]` when the
server's instructions string is non-empty. Empty string keeps the existing
JSON shape for the common no-instructions case.

## Tests

- `mcp_test.go`: round-trip JSON marshaling for `MCPServerConfig.AlwaysLoad`
  (true, false, nil-omitted) and unmarshaling.
- `protocol_test.go`: control-init body wires `alwaysLoad` through the
  `McpServerConfigs()` snapshot; SDK-MCP `initialize` response includes
  `instructions` when set and omits it when empty.
- `transport_test.go`: `--mcp-config` encoding for stdio config with
  `alwaysLoad: true`.
- `integration_test.go`: `TestIntegrationMCPAlwaysLoad` and
  `TestIntegrationMCPSDKServerInstructions` are placeholders with explicit
  `t.Skip` reasons (the CLI shape these affect isn't directly assertable from
  the SDK transport, similar to existing patterns in this file).

Out of scope: per-tool `alwaysLoad` (TS `tool({ alwaysLoad })` at
`sdk.d.ts` L5721). Brief defers it to a follow-up if needed.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -609,6 +609,22 @@ func TestIntegrationMCPTimeout(t *testing.T) {
 	t.Skip("not directly triggerable from CLI without slow stdio MCP fixture: in-process SDK MCP servers are registered separately from MCPServerConfig timeout")
 }
 
+func TestIntegrationMCPAlwaysLoad(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// TODO: Backfill when the CLI exposes prompt tool-loading state to SDK tests.
+	t.Skip("not directly assertable from CLI: alwaysLoad affects prompt composition and tool-search gating, not anything observable on the SDK transport")
+}
+
+func TestIntegrationSDKMCPInstructions(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// TODO: Backfill when the CLI exposes consumed MCP instructions to SDK tests.
+	t.Skip("not directly assertable from CLI: MCP server instructions are consumed internally by the CLI for prompt composition")
+}
+
 // TestIntegrationStopHookBlock tests that Stop hooks can block session exit
 // and reinject a new prompt using the Decision/Reason/SystemMessage fields.
 //
--- a/mcp.go
+++ b/mcp.go
@@ -14,9 +14,10 @@ import (
 //
 // Use CreateMcpServer to create a new server and AddTool to register tools.
 type McpServer struct {
-	name    string
-	version string
-	tools   map[string]*toolEntry
+	name         string
+	version      string
+	instructions string
+	tools        map[string]*toolEntry
 }
 
 // toolEntry stores tool metadata and handler.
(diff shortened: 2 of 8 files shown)


Roasbeef:
**VERDICT: APPROVE** _(orchestrator inline review — author worker bailed; transport encoding hook for `alwaysLoad` added by orchestrator before push)_

Scope matches BRIEF:
- `MCPServerConfig.AlwaysLoad *bool` with `omitempty` — distinguishes unset / explicit false / true, mirrors PR-06's `Timeout *int` rationale.
- `transport.go` writes `alwaysLoad` into the flat `--mcp-config` payload alongside `timeout`, before the type-specific switch — uniform across stdio/http/sse/socket.
- `McpServerOptions.Instructions` + private field + `Instructions()` getter — mirrors TS `SdkMcpServerConfig.instructions`.
- `handleSDKMCPMessage` initialize: extracts `result` to a separate map and conditionally sets `result["instructions"]` only when non-empty. Empty string preserves existing JSON shape for the no-instructions case.

Tests:
- Round-trip JSON marshaling for `AlwaysLoad` (true / false / nil-omitted / unmarshal) — covers the explicit-false case that the `*bool` design exists to support.
- `protocol_test.go` covers both "instructions present" and "instructions omitted when empty" paths, plus end-to-end control-init wiring of `alwaysLoad` through `McpServerConfigs()`.
- `--mcp-config` transport encoding test for stdio + alwaysLoad.
- Integration tests are `t.Skip` with reasons matching existing patterns (PR-06 set the precedent).

CI fully green across Lint/Test/Build matrix (linux/darwin/windows × amd64/arm64). Out-of-scope per-tool `alwaysLoad` from `tool({ alwaysLoad })` correctly deferred per brief.

Squash-merging.

PR #68, opened 2026-06-01
settings: add hook command `args` (exec form)

Roasbeef:
TS SDK v0.3.150 added `args?: string[]` to the `'command'`-type hook
shape in settings. When present, the CLI resolves `command` as an
executable and spawns it directly with these arguments -- no shell.
Path placeholders like `${CLAUDE_PLUGIN_ROOT}` are substituted per-element
as plain strings, so paths with quotes, `$`, or backticks never reach
a shell parser. When absent, `command` runs through a shell (bash on
POSIX, PowerShell on Windows without Git Bash).

The Go side is pure shape passthrough -- the CLI consumes `args` to
pick between exec-form and shell-form spawn; the SDK just round-trips
the field. Adds one `Args []string` field on `SettingsHook` with
`omitempty` to keep the existing JSON shape unchanged for hooks that
don't set it.

A custom `MarshalJSON` preserves an intentional non-nil empty slice as
`"args":[]` (force-exec-form-no-args), since stdlib `omitempty` would
otherwise collapse it with nil.

Part of the v0.3.150 catchup (PR 8 of 34).

See `memory/catchup-v0.3.150/PLAN.md` §"PR 08".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -625,6 +625,14 @@ func TestIntegrationSDKMCPInstructions(t *testing.T) {
 	t.Skip("not directly assertable from CLI: MCP server instructions are consumed internally by the CLI for prompt composition")
 }
 
+func TestIntegrationHookCommandArgs(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// TODO: Backfill if the CLI exposes hook subprocess spawn details to SDK tests.
+	t.Skip("not directly assertable from CLI: hook args is a settings shape consumed by the CLI's hook spawn path, not observable on the SDK transport")
+}
+
 // TestIntegrationStopHookBlock tests that Stop hooks can block session exit
 // and reinject a new prompt using the Decision/Reason/SystemMessage fields.
 //
--- a/options.go
+++ b/options.go
@@ -426,8 +426,15 @@ type SettingsHookMatcher struct {
 }
 
 type SettingsHook struct {
-	Type           string                 `json:"type"`
-	Command        string                 `json:"command,omitempty"`
+	Type    string `json:"type"`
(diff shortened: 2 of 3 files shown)


Roasbeef:
**Orchestrator review (self-approval blocker: posting as comment per project convention).**

VERDICT: APPROVE

Diff against BRIEF (`memory/catchup-v0.3.150/pr-08-hook-cmd-args/BRIEF.md`):

- ✅ `Args []string` field added to `SettingsHook` in `options.go` immediately after `Command`, with `json:"args,omitempty"` and the verbatim docstring from the BRIEF.
- ✅ Round-trip unit tests in new `options_test.go` cover all four required cases: marshal-with-args, nil-omitted, unmarshal, and the non-nil-empty-emit-`[]` shape.
- ✅ Integration test stub `TestIntegrationHookCommandArgs` follows PR-06/PR-07 skip precedent with TODO and matching skip reason.
- ✅ `go test ./...`, `go vet ./...`, `gofmt -l .` all clean locally.

Deviation from BRIEF (acceptable):

- BRIEF claimed `omitempty` on a non-nil empty slice would emit `"args":[]`. That's incorrect — `encoding/json` `omitempty` checks `len(x) == 0`, so it collapses nil and empty identically. To preserve the BRIEF's "force exec form with no args" shape, the author added a custom `MarshalJSON` that re-marshals through `map[string]interface{}` to re-inject `"args"` when non-nil. The map round-trip is safe for this struct (string / []string / map[string]string only — no numeric fields whose precision could shift). Slightly heavier than needed; a flat sentinel would be cleaner but this works and is well-tested.

Merging once all Build/Test/Lint matrix entries land green. Integration Tests already skipped (auth-gated). Will squash + delete branch.

PR #69, opened 2026-06-01
settings: add hook `continueOnBlock` field

Roasbeef:
TS SDK v0.3.150 added `continueOnBlock?: boolean` to the `'prompt'`-type
hook shape in settings. It sets the continue value for the
decision:"block" produced when the hook returns ok=false. Default
false (turn ends). Whether continue=true lets the turn proceed depends
on the event's decision:"block" semantics: on PostToolUse, the reason
is fed back to Claude and the turn continues.

The Go-side `SettingsHook` struct is a flat union over all hook
subtypes (command / prompt / agent / etc.), so the new field lives
there. Pure shape passthrough -- the CLI consumes it; the SDK just
round-trips the JSON.

Modeled as `*bool` (not `bool`) so a settings author can explicitly
set `false` to override an organization-default policy that set
`true` -- matches the precedent of `Once`, `Async`, `AsyncRewake`, and
`MCPServerConfig.AlwaysLoad` in the same package.

Part of the v0.3.150 catchup (PR 9 of 34).

See `memory/catchup-v0.3.150/PLAN.md` §"PR 09".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -633,6 +633,14 @@ func TestIntegrationHookCommandArgs(t *testing.T) {
 	t.Skip("not directly assertable from CLI: hook args is a settings shape consumed by the CLI's hook spawn path, not observable on the SDK transport")
 }
 
+func TestIntegrationHookContinueOnBlock(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// TODO: Backfill if the CLI exposes hook turn-continuation decisions to SDK tests.
+	t.Skip("not directly assertable from CLI: continueOnBlock affects the CLI's turn-continuation decision after a blocking hook, not anything observable on the SDK transport")
+}
+
 // TestIntegrationStopHookBlock tests that Stop hooks can block session exit
 // and reinject a new prompt using the Decision/Reason/SystemMessage fields.
 //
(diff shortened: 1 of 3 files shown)


Roasbeef:
**VERDICT: APPROVE** (inline orchestrator review — self-approval blocked since author == reviewer == `gh` token holder).

Diff matches BRIEF exactly:

- `ContinueOnBlock *bool` placed after `Model`, before `StatusMessage` — mirrors TS field order.
- `*bool` rationale (explicit-false-not-collapsed, matches `Once`/`Async`/`AsyncRewake`/`AlwaysLoad` precedent) — correct.
- Docstring is the TS docstring verbatim (verified against `scratch/ts-sdk-0.3.150/package/sdk.d.ts` L4218–4247).
- `MarshalJSON` from PR-08 left untouched — correct, pointer + omitempty round-trips via stdlib without going through the map.
- 4 JSON test cases: marshal-true, nil-omitted, marshal-false, unmarshal — all four matter and all are present.
- Integration `t.Skip` with TODO + reason copy matches PR-08 precedent.

CI: all green (Lint / Test / Build linux/amd64+arm64 / darwin/amd64+arm64 / windows/amd64). Integration skipped (no token in CI — expected).

Squash-merging.

PR #72, opened 2026-06-01
stream: add BackgroundTasks runtime control

Roasbeef:
In this PR, we add the `BackgroundTasks` control method on `Stream`, which backgrounds in-flight foreground tool calls (Bash commands and subagents). When `toolUseID` is non-empty, only the task started by that `tool_use` block is backgrounded; passing an empty string backgrounds all foreground tasks, equivalent to pressing Ctrl+B in the terminal. Each blocking tool call returns immediately with a "running in the background" `tool_result` and the turn continues, with the task emitting a `task_notification` once it eventually settles.

This is part of the v0.2.119 SDK catchup tracking PR 12 in the upstream TS sequence. The method maps to the `background_tasks` SDK control subtype landing in CLI v0.3.150.

## Method shape

```go
func (s *Stream) BackgroundTasks(ctx context.Context, toolUseID string) (bool, error)
```

We default the `backgrounded` response field to `true` when absent, mirroring the TS SDK's `?? true` fallback, and return `false` only when a specific `toolUseID` matched no foreground task. The existing `tool_use_id` field on `SDKControlRequestBody` is reused for the request payload; its inline comment picks up the new subtype.

## Tests

Unit coverage in `client_file_plugin_control_test.go` exercises four cases against the wire shape: with `tool_use_id`, without `tool_use_id`, an explicit `backgrounded: false` no-match response, and the absent-key default-true fallback.

```
=== RUN   TestStreamBackgroundTasksWireShape
--- PASS: TestStreamBackgroundTasksWireShape (0.00s)
    --- PASS: with_tool_use_id (0.00s)
    --- PASS: without_tool_use_id (0.00s)
    --- PASS: no_match_returns_false (0.00s)
    --- PASS: absent_backgrounded_key_defaults_to_true (0.00s)
```

A live integration test in `integration_test.go` sends a `Bash sleep` request, calls `BackgroundTasks("")` once a `tool_use` arrives, and asserts the turn completes with a backgrounded `tool_result`. It skips cleanly when the installed CLI rejects the `background_tasks` subtype, so the test slot is in place ahead of v0.3.150 rollout.

### DIFF
--- a/client.go
+++ b/client.go
@@ -908,6 +908,37 @@ func (s *Stream) StopTask(ctx context.Context, taskID string) error {
 	return err
 }
 
+// BackgroundTasks backgrounds in-flight foreground tasks (Bash commands and
+// subagents). When toolUseID is non-empty, only the task started by that
+// tool_use block is backgrounded; when empty, all foreground tasks are
+// backgrounded, equivalent to pressing Ctrl+B in the terminal.
+//
+// Each blocking tool call returns immediately with a "running in the
+// background" tool_result and the turn continues; the task keeps running
+// and emits a task_notification when it settles.
+//
+// Returns true when at least one task was backgrounded; returns false only
+// when toolUseID was non-empty and matched no foreground task.
+func (s *Stream) BackgroundTasks(
+	ctx context.Context, toolUseID string,
+) (bool, error) {
+	resp, err := s.sendSDKControlRequest(ctx, SDKControlRequestBody{
+		Subtype:   "background_tasks",
+		ToolUseID: toolUseID,
+	})
+	if err != nil {
+		return false, err
+	}
+
+	backgrounded := true
+	if v, ok := resp.Response.Response["backgrounded"]; ok {
+		if b, ok := v.(bool); ok {
+			backgrounded = b
+		}
+	}
+	return backgrounded, nil
+}
+
 // sendSDKControlRequest sends an SDK-format control request and waits for the
 // matching response.
 func (s *Stream) sendSDKControlRequest(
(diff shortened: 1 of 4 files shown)


Roasbeef:
## VERDICT: APPROVE (inline orchestrator review)

Diff matches BRIEF exactly:

- `client.go` — `Stream.BackgroundTasks` placed immediately after `StopTask`. Wire form `{subtype:"background_tasks", tool_use_id?:<id>}`, reuses the existing `SDKControlRequestBody.ToolUseID` field. Response decode does the map lookup + bool type-assert with default-true on absence, matching the TS `?? true` fallback.
- `messages.go` — single-word comment refresh on `ToolUseID` to include `background_tasks` alongside `can_use_tool/hooks`. No new field, matches the "non-goals" guidance.
- `client_file_plugin_control_test.go` — `TestStreamBackgroundTasksWireShape` with all four sub-tests from the brief: with `tool_use_id`, without (omitted from wire form), no-match returns false, absent key defaults to true. Uses existing `successSDKControlResponseWithPayload` helper — no new exported test plumbing.
- `integration_test.go` — `TestIntegrationBackgroundTasks` with the BashTool harness pattern, captures `tool_use` id, calls `BackgroundTasks(ctx, "")`, gracefully `t.Skip`s when the active CLI build predates the responder (`strings.Contains(err.Error(), "background_tasks")`).

Initial push omitted `strings` import in `integration_test.go` (typecheck-only failure under `-tags integration`); orchestrator pushed `45cc20b` to add it. Build/Test/gofmt/vet all clean locally with the fix.

Squash + delete-branch on next heartbeat once full CI is green.

PR #84, opened 2026-06-02
client: document apply_flag_settings null-value clear semantics

Roasbeef:
TS SDK v0.3.150 narrows `Stream.applyFlagSettings` to `[K in keyof Settings]?: Settings[K] | null` and documents that a `null` value clears that top-level key from the flag layer while `undefined` / a missing key is a no-op (sdk.d.ts L2170–2180).

The Go SDK is already wire-compatible — `map[string]interface{}{"k": nil}` marshals to `{"k": null}` via the stdlib. This PR refreshes the `Stream.ApplyFlagSettings` docstring to make the contract explicit and pins the wire shape with two new unit tests:

- `TestStreamApplyFlagSettingsNullValue` — single nil-valued key → JSON null.
- `TestStreamApplyFlagSettingsMixedNullValues` — mixed concrete + nil round-trip.

The `apply_flag_settings` integration sub-test inside `TestIntegrationStreamFileAndRuntime` is extended to set a flag key, then clear it with a follow-up null-valued call, both via live CLI round-trip.

Refs: TS SDK v0.3.150, `applyFlagSettings` signature L2170–2180, `SDKControlApplyFlagSettingsRequest` L2617–2622 (wire-shape unchanged).

### DIFF
--- a/client.go
+++ b/client.go
@@ -889,8 +889,17 @@ func (s *Stream) ReloadPlugins(
 	return &out, nil
 }
 
-// ApplyFlagSettings merges settings into the active flag settings layer.
-// Top-level keys are shallow-merged by the CLI across successive calls.
+// ApplyFlagSettings merges the provided settings into the flag settings layer,
+// updating the active configuration. Top-level keys are shallow-merged by the
+// CLI across successive calls and fall back to lower-precedence sources when
+// absent (an omitted key is a no-op).
+//
+// To clear a previously-set top-level key from the flag layer, pass it
+// explicitly with a nil value - this marshals to JSON null on the wire, which
+// the CLI treats as a clear-this-key signal. Passing nil as the whole settings
+// map sends an empty object (no-op for all keys).
+//
+// Only available in streaming input mode.
 func (s *Stream) ApplyFlagSettings(
 	ctx context.Context,
 	settings map[string]interface{},
--- a/client_file_plugin_control_test.go
+++ b/client_file_plugin_control_test.go
@@ -399,6 +399,46 @@ func TestStreamApplyFlagSettingsNil(t *testing.T) {
 	)
 }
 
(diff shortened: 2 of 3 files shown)


Roasbeef:
**VERDICT: APPROVE** (posted as comment due to self-PR — `gh` token is PR author).

Diff matches the BRIEF verbatim — 3 files, +61/-2:

- `client.go` — docstring-only refresh on `Stream.ApplyFlagSettings`. Calls out null-clears semantics and omit-key=no-op semantics, mirroring TS v0.3.150 sdk.d.ts L2170–2180. Body unchanged; the `if settings == nil { settings = map[string]interface{}{} }` guard stays.
- `client_file_plugin_control_test.go` — `TestStreamApplyFlagSettingsNullValue` pins the single-nil-key wire shape (`{"clearMe":null}`), and `TestStreamApplyFlagSettingsMixedNullValues` pins the mixed concrete + nil round-trip (`{"keep":"value","clear":null,"num":7}`). Both use the existing transport-snapshot style with `assert.JSONEq` on `rawWrittenSDKControlRequest`.
- `integration_test.go` — `apply_flag_settings` sub-test extended in place: empty-map baseline preserved, then set `verbose: true`, then clear with `verbose: nil`. Live CLI round-trip via the file stream.

No production API change (wire is already `Record<string, unknown>`-compatible; `nil` in `map[string]interface{}` already marshals to JSON `null`). No typed wrapper, no sentinel, no scope creep.

Local verification (Go 1.24, vendored from `/home/node/.openclaw/local/go/bin/go`):
- `go test ./...` — pass
- `go vet ./...` — clean
- `gofmt -l .` — clean
- `go vet -tags integration ./...` — clean

CI status at review time: Lint pass, 5x Build pass, Test pending, Integration skipping. Will merge once Test goes green.

PR #88, opened 2026-06-02
hooks: support suppressOriginalPrompt on UserPromptSubmit hook output

Roasbeef:
TS SDK v0.3.150 adds `suppressOriginalPrompt?: boolean` to `UserPromptSubmitHookSpecificOutput` (sdk.d.ts L5801-L5809). When a UserPromptSubmit hook blocks, setting the field tells the CLI to omit the original prompt from the block message it returns to the user -- useful when the prompt itself was the problem (PII, credentials, etc.).

- Adds `HookResult.SuppressOriginalPrompt *bool` and auto-translates it into `hookSpecificOutput.suppressOriginalPrompt` for UserPromptSubmit hooks via `buildHookResponse`.
- Uses a pointer type so nil stays absent while pointer-to-false explicitly emits `false`.
- Gated on `hookType == UserPromptSubmit` so the field does not leak onto unrelated hook outputs.
- Composes with explicit `HookResult.HookSpecificOutput` maps without clobbering existing keys, matching the existing late-merge pattern.
- Adds focused unit coverage, a skipped integration placeholder, and an `INTEGRATION-FOLLOWUPS.md` entry.

Refs: sdk.d.ts v0.3.150 L5801-L5809.

### DIFF
--- a/INTEGRATION-FOLLOWUPS.md
+++ b/INTEGRATION-FOLLOWUPS.md
@@ -3,6 +3,7 @@
 - Backfill `TestIntegrationStopHookBackgroundTasks` when the CLI test fixture can deterministically create a long-running background task.
 - Backfill `TestIntegrationStopHookSessionCrons` when the CLI test fixture can deterministically create a session cron.
 - Backfill `TestIntegrationHookTerminalSequence` when the CLI test fixture can capture terminal-escape bytes emitted by hook `terminalSequence` returns. Currently those go straight to the controlling TTY and aren't observable on the SDK transport.
+- Backfill `TestIntegrationHookSuppressOriginalPrompt` when the CLI test fixture can assert the block-message body the CLI returns when a UserPromptSubmit hook returns `suppressOriginalPrompt: true`. The omission is a CLI rendering behavior, not surfaced on the SDK transport.
 - Backfill `TestIntegrationPostToolUseUpdatedToolOutput` when the CLI test fixture can deterministically run a tool whose output a PostToolUse hook rewrites, so the model receives the rewritten value rather than the original tool response.
 - Backfill `TestIntegrationAssistantMessageSubagentFields` when the CLI test fixture can deterministically run a subagent task end-to-end.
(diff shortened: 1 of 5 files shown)


Roasbeef:
VERDICT: APPROVE

Inline orchestrator review (self-PR; formal --approve blocked by author == reviewer token).

Diff matches BRIEF verbatim across all five files:
- `options.go` — `SuppressOriginalPrompt *bool` placed between `TerminalSequence` and `UpdatedToolOutput`, docstring verbatim from BRIEF (UserPromptSubmit-only, nil vs *false vs *true semantics, sdk.d.ts L5808 citation).
- `protocol.go` — late-merge block gated on `result.SuppressOriginalPrompt != nil && hookType == string(HookTypeUserPromptSubmit)`, placed after the `WatchPaths` merge, creates the envelope if absent, inserts `suppressOriginalPrompt` alongside any existing `hookSpecificOutput` keys without clobbering.
- `protocol_test.go` — `TestBuildHookResponse_SuppressOriginalPrompt` with 5 sub-tests (emit-on-true / explicit-false-emits / nil-elides / compose-with-explicit-HSO-preserves-additionalContext / ignored-on-PreToolUse-doesn't-leak).
- `integration_test.go` — `TestIntegrationHookSuppressOriginalPrompt` t.Skip mirroring PR-25/26/27 shape with rationale about CLI rendering not being SDK-observable.
- `INTEGRATION-FOLLOWUPS.md` — backfill entry slotted into the existing cluster.

Pattern consistent with PR-09 `ContinueOnBlock` (pointer-bool semantics) and PR-26 `UpdatedToolOutput` (typed convenience field on `HookResult` auto-translating into `hookSpecificOutput`). No dedicated `UserPromptSubmitHookSpecificOutput` Go struct — out of scope per Go SDK convention (map-based HSO + typed convenience fields).

CI fully green: Lint ✓ Test ✓ Build (linux amd64/arm64, darwin amd64/arm64, windows amd64) ✓ Integration SKIPPED. mergeStateStatus CLEAN.

Approving and squash-merging.

PR #97, opened 2026-06-08
messages: add ThinkingTokensMessage system subtype (v0.3.168 deferred Phase I-2)

Roasbeef:
## Summary

Adds the `thinking_tokens` system subtype from the v0.3.168 TS SDK
(`sdk.d.ts` L3795-L3807). Reported during the redacted-thinking phase
to expose a live estimate of thinking-token usage. These values are
approximate progress, not the authoritative billed `output_tokens`
on the result event — that distinction is called out in the doc
comment so callers don't reach for this field for billing.

New `ThinkingTokensMessage` struct mirrors the adjacent
`SessionStateChangedMessage` shape: a `Type` field that round-trips
`"system"` plus `MessageType()` returning `"system"` so the SDK
interface type stays consistent across the system subtypes.

## Test plan

- [x] `gofmt -l .` empty
- [x] `go vet ./...` clean
- [x] `go test ./...` PASS
- Unit coverage: dispatch, full round-trip, zero-delta retention
  (delta is not `omitempty`)
- Integration: deliberate skip; the standard CLI run does not invoke
  a thinking-capable model through a redacted-thinking phase.
  Backfill tracked in `INTEGRATION-FOLLOWUPS.md`.

### DIFF
--- a/INTEGRATION-FOLLOWUPS.md
+++ b/INTEGRATION-FOLLOWUPS.md
@@ -10,6 +10,7 @@
 - Backfill `TestIntegrationAssistantMessageError` when the CLI test fixture can deterministically force an upstream API failure into an `assistant` event.
 - Backfill `TestIntegrationResultMessageOriginTTFT` when the CLI test fixture can deterministically emit `origin` and/or `ttft_ms` on a result event.
 - Backfill `TestIntegrationTaskLifecycleFields` when the CLI test fixture can deterministically run a Task-tool subagent (populates `subagent_type` on `task_started`/`task_progress`) and pause a running task (emits `task_updated` with `status:"paused"`).
+- Backfill `TestIntegrationThinkingTokensMessage` when the CLI test fixture can deterministically invoke a thinking-capable model through a redacted-thinking phase so `thinking_tokens` system events are emitted.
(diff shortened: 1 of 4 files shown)


Roasbeef:

- **Scope.** Single new system subtype `thinking_tokens`, struct + dispatch in `messages.go`, three focused unit tests, integration skip with INTEGRATION-FOLLOWUPS.md entry. No other surfaces touched. Matches the brief exactly.
- **Surface.** `ThinkingTokensMessage` mirrors the neighboring `SessionStateChangedMessage` shape — explicit `Type string` round-trips `"system"` and `MessageType()` returns `"system"`. Consistent with how the file already handles concrete subtypes.
- **Doc comment.** Cites `sdk.d.ts` L3795-L3807 and calls out the approximate-vs-billed distinction so callers don't reach for this field for billing math. Good.
- **Tests.** Dispatch + round-trip + zero-delta retention. The zero-delta test specifically verifies the JSON keeps `estimated_tokens_delta: 0` (not omitempty stripped), which is the right invariant for a delta field.
- **Integration.** Correctly skipped — no deterministic way to invoke a thinking-capable model with a redacted-thinking phase from the standard CLI run. INTEGRATION-FOLLOWUPS.md entry added in the same style as the existing backfill list.

Issue #95, opened 2026-06-08
Parity with TypeScript Agent SDK v0.3.168


Roasbeef:
Track parity with `@anthropic-ai/claude-agent-sdk@0.3.168`. Prior cycle landed v0.3.150 Phase A-H parity (#64-#93). This cycle rolls in two deltas:

1. **Phase I** — deferred from v0.3.150 cycle (`memory/catchup-v0.3.150/DELTA-0.3.159.md`); 9 surfaces never authored.
2. **Phase J** — new in 0.3.159 → 0.3.168; ~13 additive surfaces.
3. **Phase K** — docs refresh.

**Detected divergence:** 740-line `sdk.d.ts` diff between v0.3.150 (shipped) and v0.3.168. Purely additive — no removals.

Plan: `memory/catchup-v0.3.168/PLAN.md`. Surface catalog: `memory/catchup-v0.3.168/DELTA-159-to-168.md`.

### Phase I — deferred Phase I addendum

- [ ] PR-01 — MessageDisplay hook event family
- [ ] PR-02 — SDKThinkingTokensMessage (system/thinking_tokens)
- [ ] PR-03 — SessionStartHookSpecificOutput.reloadSkills
- [ ] PR-04 — SessionTitle on ResultMessage + SessionEnd hook input
- [ ] PR-05 — Workflows + ultracode + pluginSuggestionMarketplaces settings
- [ ] PR-06 — Marketplace skipLfs (doc + round-trip test; opaque-map already supports)
- [ ] PR-07 — Docs refresh: strictMcpConfig + effort + Options.tools + MCP timeout

### Phase J — new in 0.3.159 → 0.3.168

- [ ] PR-08 — SDKControlSuccess.pending_permission_requests (already structurally present; test + docs)
- [ ] PR-09 — SDKCommandsChangedMessage (system/commands_changed)
- [ ] PR-10 — Stop + SubagentStop hookSpecificOutput.additionalContext
- [ ] PR-11 — OnUserDialog callback + UserDialogRequest/Result + request_user_dialog dispatch
- [ ] PR-12 — reload_skills control RPC + Stream.ReloadSkills
- [ ] PR-13 — register_repo_root control request + Stream.RegisterRepoRoot
- [ ] PR-14 — SDKAssistantMessageError 'overloaded' variant
- [ ] PR-15 — MessageOriginKind 'auto-continuation' variant
[...]

PR #111, opened 2026-06-08
messages: add ResultMessage timing fields (v0.3.168 Phase J-9)

Roasbeef:
Tracks four optional `SDKResultSuccess` timing fields added upstream in
`@anthropic-ai/claude-agent-sdk@0.3.168` (sdk.d.ts L3566-L3569).

| TS field | Go field |
|----------|----------|
| `ttft_stream_ms?: number` | `TTFTStreamMs *int64` |
| `time_to_request_ms?: number` | `TimeToRequestMs *int64` |
| `time_to_request_from_spawn_ms?: number` | `TimeToRequestFromSpawnMs *int64` |
| `warm_spare_claimed?: boolean` | `WarmSpareClaimed *bool` |

All four are placed between `TTFTMs` and `IsError` to match the upstream
field order. Pointer types preserve the absent-vs-zero distinction under
`omitempty` — important for `warm_spare_claimed: false`, which has to
round-trip as an explicit `false`, not be elided.

Unit coverage in `messages_test.go`: parse, omitempty elision, explicit-false
round-trip, full round-trip with all four populated.

Integration is skipped: these are populated by CLI-internal spawn-pool
timing instrumentation and aren't reachable from the standard integration
run. An INTEGRATION-FOLLOWUPS entry tracks backfilling once the CLI test
fixture can deterministically emit them.

Part of v0.3.168 Phase J. Refs #95.

### DIFF
--- a/INTEGRATION-FOLLOWUPS.md
+++ b/INTEGRATION-FOLLOWUPS.md
@@ -11,6 +11,7 @@
 - Backfill `TestIntegrationUserMessageSubagentFields` when the CLI test fixture can deterministically run a Task-tool subagent whose tool-result user message round-trips `subagent_type` + `task_description`.
 - Backfill `TestIntegrationAssistantMessageError` when the CLI test fixture can deterministically force an upstream API failure into an `assistant` event.
 - Backfill `TestIntegrationResultMessageOriginTTFT` when the CLI test fixture can deterministically emit `origin` and/or `ttft_ms` on a result event.
+- Backfill `TestIntegrationResultMessageTimingFields` when the CLI test fixture can deterministically emit spawn-pool timing instrumentation (`ttft_stream_ms`, `time_to_request_ms`, `time_to_request_from_spawn_ms`, `warm_spare_claimed`) on a result event.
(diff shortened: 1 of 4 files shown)


Roasbeef:
Diff matches the brief:

- `ResultMessage` gains `TTFTStreamMs`, `TimeToRequestMs`, `TimeToRequestFromSpawnMs` (`*int64`) and `WarmSpareClaimed` (`*bool`), placed between `TTFTMs` and `IsError` — same order as `SDKResultSuccess` at sdk.d.ts L3566-L3569.
- Pointer types are the right call for `warm_spare_claimed: false` to survive `omitempty` round-trips. `TestResultMessageTimingFieldsExplicitFalse` locks that in.
- Coverage: parse path, omitempty elision, explicit-false round-trip, full populated round-trip — the four shapes we actually care about.
- Integration skip is correct: spawn-pool timing is CLI-internal instrumentation, not driveable from a normal session. INTEGRATION-FOLLOWUPS entry tracks the backfill condition.

APPROVED.

Issue #125, opened 2026-06-22
Parity with TypeScript Agent SDK v0.3.185


Roasbeef:
Track Go SDK parity with `@anthropic-ai/claude-agent-sdk@0.3.185` (prior shipped: v0.3.177, #116–#124).

Delta: 2 new system message subtypes, ~7 new fields, 1 optional transport method, Settings parity, docs. See `memory/catchup-v0.3.185/DELTA-177-to-185.md`.

### Planned PRs

- [ ] 01 — `SDKWorkerShuttingDownMessage` (`system`/`worker_shutting_down`)
- [ ] 02 — `SDKInformationalMessage` (`system`/`informational`)
- [ ] 03 — `set_max_thinking_tokens` gains `thinking_display`
- [ ] 04 — `RateLimitInfo`: `credits_required` / `canUserPurchaseCredits` / `hasChargeableSavedPaymentMethod`
- [ ] 05 — `ResultMessage` (success): `time_origin_ms`
- [ ] 06 — `UserMessage.senderTaskId`
- [ ] 07 — Settings parity: `attribution.sessionUrl`, `disableClaudeAiConnectors`, `sandbox.allowAppleEvents`
- [ ] 08 — optional `WaitForExit` capability on `Transport`
- [ ] 09 — docs refresh + parity bump

### Gated / deferred

- `SDKControlSetMcpPermissionModeOverrideRequest` — referenced in the control-request union but no exported type body / no wire literal in `sdk.mjs`. Nothing to model; deferred until upstream emits a definition.

### Removed upstream (no-op for Go)

- `ConnectRemoteControl{Error,Options,Result}`, `InboundPrompt` — `@alpha` bridge types, never ported.

Roasbeef:
v0.3.185 parity complete. All 9 PRs squash-merged to main (`2d40ee9`):

- #126 worker_shutting_down system message
- #129 informational system message
- #130 thinking_display on set_max_thinking_tokens
- #131 RateLimitInfo credit-exhaustion fields
- #132 ResultMessage.time_origin_ms
- #133 UserMessage.senderTaskId
- #127 Settings parity (disableClaudeAiConnectors, attribution.sessionUrl, sandbox.allowAppleEvents)
- #128 transport ExitWaiter/WaitForExit
- #134 docs refresh + doc-only mirrors

Build/Lint/Test green on every matrix for all PRs.

[...]

PR #126, opened 2026-06-22
messages: add worker_shutting_down system message

Roasbeef:
Ports `SDKWorkerShuttingDownMessage` from TS SDK v0.3.185 (sdk.d.ts L4205–4218). Part of #125.

- New `WorkerShuttingDownMessage` (`type: system`, `subtype: worker_shutting_down`) carrying `reason`, `uuid`, `session_id`.
- Dispatch case in `ParseMessage`'s system-subtype switch.
- Docstring captures the upstream contract: emitted only on opt-in graceful teardown with a reason; absence is not a dead-host signal; it lands in the durable event stream so treat it as a live-tail signal, not a session-lifetime fact (CC-2656).
- Unit test round-trips a canonical payload through `ParseMessage`.

Integration: skipped — graceful-teardown emit is not deterministically triggerable from the CLI.

See memory/catchup-v0.3.185/PLAN.md §"PR 01".


Roasbeef:
Self-review (token = author, formal approve blocked): additive, mirrors the cited sdk.d.ts surface, unit coverage present, CI green on all build/lint/test matrices. VERDICT: APPROVE.

PR #128, opened 2026-06-22
transport: add optional WaitForExit capability

Roasbeef:
Ports the optional `Transport.waitForExit` from TS SDK v0.3.185 (sdk.d.ts L6418). Part of #125.

- New `ExitWaiter` capability interface — `WaitForExit(ctx) error`. Kept off the core `Transport` interface so external `Transport` implementers don't break; callers type-assert.
- `*SubprocessTransport` implements it: blocks until the CLI subprocess exits or ctx is done.
- The exit wait is now shared between `Close` and `WaitForExit` via a `sync.Once` latch, so the underlying `os/exec.Cmd` is reaped exactly once and concurrent callers all observe the same result.
- Tests cover the exit-error path, idempotency, and context cancellation (blocking runner).

Mirrors the upstream intent: cleanup awaits exit (bounded) so teardown doesn't resolve while the child is still draining the stdin EOF that Close sent.

See memory/catchup-v0.3.185/PLAN.md §"PR 08".

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -60,6 +60,25 @@ type Transport interface {
 	IsReady() bool
 }
 
+// ExitWaiter is an optional capability implemented by transports backed by a
+// subprocess whose exit can be awaited. Network and in-process transports do
+// not implement it. Mirrors the optional Transport.waitForExit in the TS SDK
+// (sdk.d.ts v0.3.185 L6418): cleanup awaits it (bounded) so teardown does not
+// resolve while the child is still draining the stdin EOF that Close sent.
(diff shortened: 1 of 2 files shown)


Roasbeef:
Self-review (token = author, formal approve blocked): optional capability kept off the core Transport interface so external implementers don't break; shared sync.Once exit latch avoids double-Wait on os/exec.Cmd; tests cover exit-error, idempotency, and ctx-cancel. CI green on all matrices. VERDICT: APPROVE.

PR #130, opened 2026-06-22
protocol: add thinking_display to set_max_thinking_tokens

Roasbeef:
Ports the `thinking_display` parameter added to `set_max_thinking_tokens` in TS SDK v0.3.185 (sdk.d.ts L2229 method, L3432 control body). Part of #125.

- `SDKControlRequestBody.ThinkingDisplay *ThinkingDisplayOverride` — a wrapper over the existing `ThinkingDisplay` type that distinguishes three wire states: omitted (keep session-start display), a mode value, and explicit `null` (clear to API default).
- `Stream.SetMaxThinkingTokens` gains a variadic `...SetMaxThinkingTokensOption` (backward-compatible) with `WithThinkingDisplay(mode)` and `WithThinkingDisplayAPIDefault()`.
- Reuses the existing `ThinkingDisplay` / `ThinkingDisplaySummarized` / `ThinkingDisplayOmitted` rather than redeclaring.
- Unit tests round-trip the control body for all three states (omitted / value / explicit null).

Integration: covered by the control-request round-trip unit tests, matching the existing `SetMaxThinkingTokens` coverage pattern (the base method ships unit-only; the display override is not separately observable from the CLI).

See memory/catchup-v0.3.185/PLAN.md §"PR 03".

### DIFF
--- a/client.go
+++ b/client.go
@@ -783,13 +783,52 @@ func (s *Stream) SetModel(ctx context.Context, model string) error {
 	return err
 }
 
+// setMaxThinkingTokensOptions accumulates optional set_max_thinking_tokens
+// parameters.
+type setMaxThinkingTokensOptions struct {
+	display *ThinkingDisplayOverride
+}
+
+// SetMaxThinkingTokensOption configures Stream.SetMaxThinkingTokens.
+type SetMaxThinkingTokensOption func(*setMaxThinkingTokensOptions)
+
+// WithThinkingDisplay sets the thinking display mode for the rest of the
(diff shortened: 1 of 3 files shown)


Roasbeef:
Self-review (token = author, formal approve blocked): three-state wire field modeled with a dedicated wrapper (omit/value/null) reusing the existing ThinkingDisplay enum; variadic option keeps the method signature backward-compatible; round-trip tests cover all three states. CI green on all matrices. VERDICT: APPROVE.

PR #136, opened 2026-06-29
client: add SetMcpPermissionModeOverride control request

Roasbeef:
Ungates the `set_mcp_permission_mode_override` control request, which was deferred in the v0.3.185 cycle (referenced in the `SDKControlRequestInner` union but with no wire shape). TS SDK v0.3.195 exposes it as `Query.setMcpPermissionModeOverride(serverName, mode)` (sdk.d.ts L2211-2228; wire shape from sdk.mjs: `{subtype:"set_mcp_permission_mode_override", serverName, mode}`).

See `memory/catchup-v0.3.195/PLAN.md` §"PR 01".

- New `Stream.SetMcpPermissionModeOverride(ctx, serverName, mode *McpPermissionOverrideMode) (McpPermissionModeOverrideResult, error)`. `mode` is a tristate: `McpPermissionOverrideModeDefault` / `McpPermissionOverrideModeAuto` send the string; a nil pointer serializes as explicit wire `null` to clear the override. Tighten-only — can never widen privilege.
- The shared `SDKControlRequestBody.Mode` field is `omitempty` and cannot emit an explicit null, so this subtype uses a dedicated wire envelope. Factored the send/await core of `sendSDKControlRequest` into `dispatchControlRequest`, which both paths share.
- Result carries the optional `warning` (typo detection: set when `serverName` matches no known MCP server; the override is stored regardless).
- Unit tests: tristate `mode` round-trip (default/auto/null) + `warning` parse. Integration test exercises the live control request and skips cleanly on CLIs predating the subtype (the installed CLI 2.1.177 returns `Unsupported control request subtype`, confirming the wire string is correct).

### DIFF
--- a/client.go
+++ b/client.go
@@ -783,6 +783,80 @@ func (s *Stream) SetModel(ctx context.Context, model string) error {
 	return err
 }
 
+// McpPermissionOverrideMode is a per-MCP-server permission-mode override. The
+// override is tighten-only: it applies only when the session mode would
+// already auto-allow (bypassPermissions/auto), so it can never widen
+// privilege.
+type McpPermissionOverrideMode string
+
+const (
+	// McpPermissionOverrideModeDefault forces per-action permission prompts
+	// for the server.
+	McpPermissionOverrideModeDefault McpPermissionOverrideMode = "default"
+	// McpPermissionOverrideModeAuto routes the server through the auto-mode
+	// classifier.
+	McpPermissionOverrideModeAuto McpPermissionOverrideMode = "auto"
+)
+
+// McpPermissionModeOverrideResult is the result of SetMcpPermissionModeOverride.
+type McpPermissionModeOverrideResult struct {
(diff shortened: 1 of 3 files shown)


Roasbeef:

VERDICT: APPROVE

- Wire shape matches sdk.mjs (`{subtype, serverName, mode}`); tristate `mode` correctly serializes nil → explicit JSON `null` via a dedicated envelope rather than overloading the omitempty shared `Mode` field. Sound — the shared body cannot emit explicit null.
- `dispatchControlRequest` refactor preserves the register-before-write ordering and error/cancel handling of the original `sendSDKControlRequest`; no behavior change for existing callers.
- Tighten-only semantics and the `warning` typo-detection field documented and tested. Integration test skips cleanly on CLIs predating the subtype (verified live: CLI 2.1.177 returns `Unsupported control request subtype`, confirming the wire literal).

Ungates the v0.3.185 GATED item. Merging.

PR #138, opened 2026-06-29
protocol: add Reinitialize to re-send initialize

Roasbeef:
Mirrors `Query.reinitialize()` added in TS SDK v0.3.195 (sdk.d.ts L2289-2306). In the TS SDK, `reinitialize()` is `() => this.initialize()` — it forces a fresh `initialize` control request instead of returning the cached first-connect result.

See `memory/catchup-v0.3.195/PLAN.md` §"PR 02".

- `Stream.Reinitialize(ctx) (*SDKControlInitializeResponse, error)` re-sends the `initialize` control request to the running CLI and returns the fresh response, refreshing the cached one.
- `Protocol.Initialize` keeps its idempotent guard; the request-building body is factored into `doInitialize`, which both `Initialize` and the new `Protocol.Reinitialize` (guard-bypassing) call.
- Use after a transport gap so the CLI redelivers any `can_use_tool` / `request_user_dialog` requests the loop is still blocked on; callbacks should be idempotent per request_id (documented on the method).
- Unit test: pre-marks the session initialized, then asserts Reinitialize still issues a fresh request and refreshes the cache (distinguishable `output_style`). Integration test passes live against the installed CLI (re-sends initialize and gets a fresh response).

### DIFF
--- a/client.go
+++ b/client.go
@@ -1244,6 +1244,24 @@ func (s *Stream) InitializationResult() (*SDKControlInitializeResponse, error) {
 	return cloneInitializeResponse(initResp), nil
 }
 
+// Reinitialize re-sends the initialize control request to the running CLI and
+// returns the fresh response. Unlike InitializationResult, which returns the
+// cached first-connect result, this always issues a new request. Use it after
+// a transport gap so the CLI redelivers any can_use_tool / request_user_dialog
+// control requests the loop is still blocked on; the SDK dedupes in-flight
+// request_ids, but callbacks should be idempotent per request_id since a
+// request whose response was lost in the gap is dispatched again.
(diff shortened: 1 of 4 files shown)


Roasbeef:

VERDICT: APPROVE

- Mirrors TS `reinitialize() => this.initialize()`: forces a fresh `initialize` past the idempotent guard. Clean factoring — `Initialize` keeps its no-op guard, `doInitialize` holds the request build, `Reinitialize` bypasses the guard and refreshes the cache.
- No behavior change for existing `Initialize`/`Connect` callers.
- Idempotency caveat for redelivered request_ids documented on the public method.
- Unit test proves guard-bypass + cache refresh; integration test passes live against the installed CLI.

Merging.

PR #141, opened 2026-06-29
messages: add RefusedUserMessageUUID to refusal fallback

Roasbeef:
Adds the `refused_user_message_uuid` field to `ModelRefusalFallbackMessage`, new in TS SDK v0.3.195 (sdk.d.ts L3783).

See `memory/catchup-v0.3.195/PLAN.md` §"PR 04".

- `RefusedUserMessageUUID *string` (`refused_user_message_uuid,omitempty`) — UUID of the user message the refused request was for; the rewind target and composer prefill for edit-and-retry. nil when the refused turn was not human-authored (background task / auto-continuation) or cannot be identified; absent from older CLIs.
- Extends the existing `TestParseMessageModelRefusalFallback` cases: full-payload asserts the value, older-CLI asserts nil.
- No integration test (struct field; the sibling `model_refusal_no_fallback` slot already tracks the refusal path).

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -944,6 +944,13 @@ type ModelRefusalFallbackMessage struct {
 	// Absent when emitted by an older CLI.
 	RetractedMessageUUIDs []string `json:"retracted_message_uuids,omitempty"`
 
+	// RefusedUserMessageUUID is the UUID of the user message the refused
+	// request was for — the rewind target and composer prefill for
+	// edit-and-retry. nil when the refused turn was not human-authored (e.g. a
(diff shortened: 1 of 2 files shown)


Roasbeef:

VERDICT: APPROVE

- `refused_user_message_uuid` added as `*string` omitempty, mirroring sdk.d.ts v0.3.195. Semantics (rewind target / edit-and-retry prefill; nil for non-human-authored turns; absent on older CLIs) documented.
- Extends the existing fallback test both ways (value present + nil). No new public type, struct field only.

Merging.

PR #144, opened 2026-06-29
Add WithExistingAuth option

tylergannon:
## Summary
- add `WithExistingAuth()` to let SDK callers explicitly use Claude Code's existing TUI login state
- scrub known Claude auth token env vars from the subprocess environment when existing auth is enabled
- document the new auth path in the README

## Verification
- `go test ./...`
- external throwaway proof module under `/tmp/existing-auth-proof`, importing this checkout via `replace`, run with both token env vars removed:
  - `env -u ANTHROPIC_API_KEY -u CLAUDE_CODE_OAUTH_TOKEN go run .`

The external proof used `WithExistingAuth()`, `WithCwd()`, and bypass permissions to ask Claude Code to read `input.txt` and create `output.txt`; the Go harness then verified the file contents.

Live proof output:

```text
workdir=/tmp/existing-auth-tool-proof-3204693270
result=success
assistant=I'll read input.txt first, then create output.txt with the specified content.existing auth tool proof: ALPHA BETA 42
output.txt=existing auth tool proof: ALPHA BETA 42
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ go get github.com/roasbeef/claude-agent-sdk-go
 Requirements:
 - Go 1.23+ (for `iter.Seq`)
 - Claude Code CLI: `npm install -g @anthropic-ai/claude-code`
-- `ANTHROPIC_API_KEY` or `CLAUDE_CODE_OAUTH_TOKEN` in your environment
+- Either `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` in your environment,
+  or a completed Claude Code TUI login with `WithExistingAuth()`
 
 ## Quick Start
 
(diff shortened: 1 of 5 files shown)


tylergannon:
Retracting this PR because a live proof showed the unmodified upstream SDK already uses Claude Code's existing on-disk login when both documented token variables are absent.

Proof: pinned upstream efdbecd88a98, Claude Code 2.1.206, ran with `env -u ANTHROPIC_API_KEY -u CLAUDE_CODE_OAUTH_TOKEN go run .`; the SDK returned `subtype=success` with a session ID, Claude used file tools, and the Go host independently verified the created file.

`WithExistingAuth` would still defensively scrub conflicting token variables, but it is not required for ordinary existing-login operation, which was the motivation for this PR.

Issue #154, opened 2026-07-13
Parity with TypeScript Agent SDK v0.3.207


Roasbeef:
Weekly TS→Go parity catchup. Prior shipped: v0.3.201 (#146–#153).

The v0.3.201 cycle deferred four messages as "type-only, no shipped wire shape." v0.3.207 concretizes all four, plus small parity fields and one behavioral change (interrupt receipt).

## PRs

- [ ] 01 — `active_goal` message (new top-level type; value obj | null)
- [ ] 02 — `background_tasks_changed` system message (REPLACE-semantics task set)
- [ ] 03 — `control_request_progress` system message
- [ ] 04 — `conversation_reset` message (new top-level type)
- [ ] 05 — `capabilities` field on system/init (feature-detection, open set)
- [ ] 06 — interrupt receipt (`interrupt_receipt_v1`: `still_queued` uuids surviving an interrupt)
- [ ] 07 — 6 new `TerminalReason` values
- [ ] 08 — `MessageOrigin` peer `body` field
- [ ] 09 — `SessionMessage.parent_agent_id`

## Out of scope (Go doesn't model the surface)
mcp_call staging, tool input/output schemas (ScheduleWakeup/Artifact/Worktree/ExitPlanMode/ArtifactOutput), assistant API `usage` sub-fields, get_plan/get_workspace_diff outbound requests.

Roasbeef:
All 10 PRs merged; `main` green (build matrix + Lint + Test; gofmt/vet/test clean post-merge).

- #155 — active_goal message
- #156 — background_tasks_changed message
- #157 — control_request_progress message
- #158 — conversation_reset message
- #159 — system/init capabilities
- #160 — interrupt receipt (still_queued / InterruptWithReceipt)
- #161 — six new TerminalReason values
- #162 — MessageOrigin peer body
- #163 — SessionMessage parent_agent_id
- #164 — docs refresh

The four newly-concretized messages (active_goal, conversation_reset, control_request_progress, background_tasks_changed) were the headline — the first three were deferred as type-only in the v0.3.201 cycle and got concrete wire shapes in v0.3.207.

No follow-up integration issues. Out-of-scope surfaces (mcp_call staging, tool-schema churn, assistant usage API sub-fields) documented in the README parity section.

PR #155, opened 2026-07-13
messages: model active_goal message (v0.3.207)

Roasbeef:

TS SDK v0.3.207 gives `SDKActiveGoalMessage` a concrete wire shape (`sdk.d.ts` L2758) — the v0.3.201 cycle deferred it as type-only with no shipped shape.

Adds the top-level `active_goal` message: `Value *ActiveGoalValue` is nil when the goal is cleared, otherwise carries `condition`, `iterations`, `set_at`, `tokens_at_start`, and optional `last_reason`. Parse case + round-trip and cleared-value tests.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -483,6 +483,29 @@ type PromptSuggestionMessage struct {
 // MessageType implements Message.
 func (m PromptSuggestionMessage) MessageType() string { return "prompt_suggestion" }
 
+// ActiveGoalMessage reports the state of the session's /goal Stop hook: it fires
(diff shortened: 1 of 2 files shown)


Roasbeef:
Self-review (orchestrator, per catchup process — gh token is the PR author so no formal approval): matches `sdk.d.ts` v0.3.207 L2758. `Value` is a pointer so a cleared goal (`value:null`) is distinguishable from a zero goal; `last_reason` omitempty. Parse case + round-trip and cleared tests green; build matrix + Lint + Test pass. VERDICT: APPROVE.

PR #160, opened 2026-07-13
client: return interrupt receipt with still_queued uuids (v0.3.207)

Roasbeef:

TS SDK v0.3.207: `interrupt()` now resolves to `SDKControlInterruptResponse` (`sdk.d.ts` L3401) carrying `still_queued` — the uuids of async user messages that survive the interrupt and will still run unless cancelled first — on CLIs advertising the `interrupt_receipt_v1` capability (see `SystemMessage.Capabilities`, PR-05). Older CLIs send an empty success response.

Adds `InterruptWithReceipt(ctx) (*InterruptReceipt, error)` returning `StillQueued`. `Interrupt(ctx) error` stays as a thin wrapper that discards the receipt — no breaking signature change. Older CLIs yield a non-nil receipt with an empty `StillQueued`. Tests cover both the populated receipt and the empty-response path.

### DIFF
--- a/client.go
+++ b/client.go
@@ -755,11 +755,51 @@ func (s *Stream) handleSends() {
 
 // Interrupt sends an interrupt signal to stop the current generation.
 // It blocks until the CLI acknowledges the request or returns an error.
+//
+// Use InterruptWithReceipt instead when you need the interrupt receipt (the
+// uuids of async user messages that survive the interrupt); this method
+// discards it.
 func (s *Stream) Interrupt(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


Roasbeef:
Self-review (orchestrator; gh token authors the PR). Non-breaking: Interrupt keeps its `error` signature and delegates to InterruptWithReceipt. Receipt decode guards on a non-empty response map, so older-CLI empty responses yield an empty StillQueued rather than an error. Both paths tested; full suite + build matrix + Lint + Test green. VERDICT: APPROVE.

PR #162, opened 2026-07-13
messages: add MessageOrigin peer body field (v0.3.207)

Roasbeef:

TS SDK v0.3.207 adds `body?: string` to the peer `MessageOrigin` kind (`sdk.d.ts` ~L3920) — the decoded message body with the peer envelope stripped, byte-exact with what the model sees, present only when the turn is exactly one harness-formed envelope. (`name` on peer is already modeled as `MessageOrigin.Name`.)

Adds `MessageOrigin.Body` (omitempty) and extends the origin-peer parse test.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -230,6 +230,11 @@ type MessageOrigin struct {
 	// SenderTaskID is the observer's task id; set only for the
 	// "observer" kind (sdk.d.ts v0.3.201).
 	SenderTaskID string `json:"senderTaskId,omitempty"`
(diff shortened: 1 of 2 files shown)


Roasbeef:
Self-review (orchestrator; gh token authors the PR). Matches `sdk.d.ts` v0.3.207 peer origin `body`; omitempty keeps non-peer origins clean. Origin-peer parse test extended and green; build matrix + Lint + Test pass. VERDICT: APPROVE.

PR #163, opened 2026-07-13
sessions: add SessionMessage parent_agent_id (v0.3.207)

Roasbeef:

TS SDK v0.3.207 adds `parent_agent_id: string | null` to session transcript messages (`sdk.d.ts` L4593, `SessionMessage`) — the agentId of the subagent that spawned this subagent, nil for a depth-1 subagent (spawned by the main loop), the main session, or sessions whose metadata lacks the field.

Adds `SessionMessage.ParentAgentID *string`, populated in `readSessionMessages` mirroring `parent_tool_use_id`. Test covers both a set value and the null case.

### DIFF
--- a/sessions.go
+++ b/sessions.go
@@ -36,6 +36,11 @@ type SessionMessage struct {
 	SessionID       string          `json:"session_id"`
 	Message         json.RawMessage `json:"message"`
 	ParentToolUseID *string         `json:"parent_tool_use_id"`
+	// ParentAgentID is the agentId of the subagent that spawned this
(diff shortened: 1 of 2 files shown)


Roasbeef:
Self-review (orchestrator; gh token authors the PR). Matches `sdk.d.ts` v0.3.207 SessionMessage L4593. Populated from the transcript map mirroring parent_tool_use_id; null decodes to nil. Set/null test green; build matrix + Lint + Test pass. VERDICT: APPROVE.

Issue #167, opened 2026-07-20
Parity with TypeScript Agent SDK v0.3.215


Roasbeef:
Tracks the weekly catchup from TS SDK **v0.3.207** (last shipped) to **v0.3.215**.

Diff is small and mechanical (sdk.d.ts field/enum additions; sdk-tools.d.ts churn deferred as usual).

### Modeled surfaces — one PR each
- [ ] 01 `apiProvider` += `anthropicGoogleCloud` (query_types.go)
- [ ] 02 SDKAssistantMessage += `resumed_from_incomplete_thinking`, `aborted`, `timestamp` (messages.go)
- [ ] 03 PluginInfo += `version` — init system msg + reload_plugins response (query_types.go)
- [ ] 04 Permission request += `matched_ask_rule`, `suppress_always_allow_rule` (options.go, protocol.go)
- [ ] 05 MessageOrigin task-notification += `subkind: 'scheduled-trigger'` (messages.go)
- [ ] 06 SDKToolProgressMessage += `heartbeat`, `subagent_type`, `subagent_retry` (messages.go)
- [ ] 07 UserPromptSubmit += `source`; SessionStart source += `fork` (options.go)
- [ ] 08 Settings += `processWrapper`, `feedbackDrafts`, `vimInsertModeRemaps` (options.go)
- [ ] 09 README parity refresh → v0.3.215 + document deferred (README.md)

### Deferred (no matching Go surface; documented in README parity)
- sdk-tools.d.ts churn: new tools RefreshMcpTools / SendFeedback / ProposeSkills + tool field additions (curated-subset deferral)
- Alpha usage/policy prefix consts (ORG_POLICY_LIMIT_PREFIXES, USAGE_LIMIT_ERROR_PREFIXES, USAGE_TRANSITION_PREFIXES, USAGE_WARNING_PREFIXES)
- `signalCode?` on process-exit transport type
- `set_max_thinking_tokens` optional / `effortLevel: 'max'` — already representable in Go
- Doc-URL migrations + setMcpServers plugin-exemption + mcp_call on-demand notes

Roasbeef:
v0.3.215 catchup complete. All 9 PRs squash-merged to main; CI green across every build matrix entry (integration skipped on PR checks). Post-merge `go build/vet ./...`, `gofmt -l .`, and `go test ./...` all clean on main.

Merged PRs:
- #168 apiProvider anthropicGoogleCloud
[...]

PR #169, opened 2026-07-20
messages: add assistant wrapper fields for v0.3.215

Roasbeef:
## What

TS SDK v0.3.215 adds three wrapper-level fields to `SDKAssistantMessage`. Mirror them on `AssistantMessage`:

- **`resumed_from_incomplete_thinking`** — this turn continued the preceding truncated assistant turn inside its trailing signed thinking block (max-output-tokens recovery). Thinking signatures are cumulative over that preceding thinking-only turn, so history replayed through the bridge must carry this flag back for the normalizer to keep the run's prefix on the wire. Wrapper-level sibling — never inside `message.content` — so it is not replayed to the model.
- **`aborted`** — the message was truncated by an interrupt/abort before the stream completed; `stop_reason` was never received and content may end mid-word.
- **`timestamp`** — ISO-8601 completion time of this content block on the originating process. Display-only (originating host clock); not an ordering key. Older emitters omit it.

All three are additive, `omitempty`, and populated by standard unmarshal — no parse-path change.

## Tests

Extended the existing `TestParseMessageAssistantMessageSubagentFields`, `TestAssistantMessageFieldsOmitEmpty`, and `TestAssistantMessageJSONRoundTrip` to cover the new fields (parse, omitempty, round-trip).

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -103,6 +103,22 @@ type AssistantMessage struct {
 	// audit record for the turn. Absent when emitted by an older CLI.
 	Supersedes []string `json:"supersedes,omitempty"`
 
+	// ResumedFromIncompleteThinking is true when this turn continued the
+	// preceding truncated assistant turn inside its trailing signed thinking
+	// block (max-output-tokens recovery). The thinking signatures are
+	// cumulative over that preceding thinking-only turn, so history replayed
+	// through the bridge must carry this flag back for the normalizer to keep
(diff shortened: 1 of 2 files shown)


Roasbeef:

Three additive omitempty wrapper fields mirroring the TS SDKAssistantMessage additions exactly; populated by standard unmarshal, no parse-path change. Existing round-trip/omitempty tests extended. CI green across matrix; Integration skipped as expected.

PR #170, opened 2026-07-20
query: add version to PluginInfo

Roasbeef:
## What

TS SDK v0.3.215 adds an optional `version` to the plugin-info shape in two places:

- `SDKControlReloadPluginsResponse.plugins[]` (the `reload_plugins` control response)
- the `initialize` system message `plugins[]`

Both are backed by Go's `PluginInfo`, so a single `omitempty` `Version` field covers both call sites.

The value is the plugin's version as declared in its `plugin.json` manifest, emitted verbatim — **plugin-author-controlled, so validate before trusting**. Omitted when the manifest declares no version.

## Tests

Extended the existing `reload_plugins` round-trip test to carry a `version` on the returned plugin.

### DIFF
--- a/client_file_plugin_control_test.go
+++ b/client_file_plugin_control_test.go
@@ -301,9 +301,10 @@ func TestStreamReloadPluginsParsesResponse(t *testing.T) {
 				},
 				"plugins": []interface{}{
 					map[string]interface{}{
-						"name":   "local",
-						"path":   "/tmp/plugin",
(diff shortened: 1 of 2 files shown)


Roasbeef:

Single omitempty field on PluginInfo, which backs both the reload_plugins response and the init system message — covers both TS call sites. Reload round-trip test extended. CI green; Integration skipped as expected.

PR #171, opened 2026-07-20
protocol: parse ask-rule fields on permission requests

Roasbeef:
## What

TS SDK v0.3.215 adds two fields to the `can_use_tool` control request (`SDKControlPermissionRequest`). Surface both on `PermissionContext` and parse them in `handlePermissionRequest`:

- **`suppress_always_allow_rule`** — the dialog must not offer the persistent "don't ask again" affordance for this ask, because accepting it would write a whole-tool allow rule broader than the ask's own verb. Hosts rendering approve options should omit any persistent-rule row when set.
- **`matched_ask_rule`** (`{source, tool_name, rule_content?}`) — a user-configured ask rule (`permissions.ask`) forced this prompt, but the ask still carries the tool's own decision reason, so the rule rides here instead of a `"rule"` `decision_reason_type`. Hosts making policy on the decision reason (auto-deny a safetyCheck) or running host-side auto-approval should treat asks carrying this as rule-forced. **Producer-authored and render-unsafe** like a decision reason — sanitize before display.

Also widened the `RequiresUserInteraction` doc to cover the v0.3.215 `localDisplayOnly` case (its consent disclosure cannot ride the wire; only the local dialog renders it).

## Tests

Added `TestProtocolHandlePermissionRequestAskRule` (fields parsed into context; absent fields leave zero values).

### DIFF
--- a/options.go
+++ b/options.go
@@ -1429,6 +1429,21 @@ type ToolPermissionRequest struct {
 	Context   PermissionContext
 }
 
+// MatchedAskRule describes the user-configured ask rule (permissions.ask) that
+// forced a permission prompt whose ask nonetheless carries the tool's own
+// decision reason. When set, the ask-rule substitution kept the richer
+// tool-minted ask, so the rule rides here instead of a "rule"
+// decision_reason_type. Hosts making policy on the decision reason (e.g.
+// auto-deny a safetyCheck) or running host-side auto-approval should treat asks
+// carrying this as rule-forced: the user's stated intent is a human prompt. The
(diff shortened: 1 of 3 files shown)


Roasbeef:

Two additive permission-request fields parsed defensively (type-asserted, nil on absent/malformed) and surfaced on PermissionContext, mirroring the TS SDKControlPermissionRequest additions. matched_ask_rule doc correctly flags the render-unsafe/sanitize requirement. New test covers both present and absent paths. CI green; Integration skipped as expected.

PR #172, opened 2026-07-20
messages: add scheduled-trigger subkind to MessageOrigin

Roasbeef:
## What

TS SDK v0.3.215 adds an optional `subkind` to the `task-notification` variant of `SDKMessageOrigin`, with a single value `'scheduled-trigger'`:

> Present when the delivery is the fired stored prompt of a scheduled task/routine (stamped from server-asserted provenance; the schedule attests storage, not authorship). The harness frames it as the session's assigned task instead of the generic background-notification frame. Absent on webhook, PR-steward, plugin, and background-event deliveries.

Adds `MessageOrigin.Subkind` (omitempty) and the `MessageOriginSubkind` type with the `MessageOriginSubkindScheduledTrigger` constant.

## Tests

Added round-trip coverage for a scheduled-trigger origin and an omitempty check for a plain task-notification origin.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -259,7 +259,24 @@ type MessageOrigin struct {
 	// only when the turn is exactly one harness-formed envelope; render it
 	// instead of re-parsing the message text (sdk.d.ts v0.3.207).
 	Body string `json:"body,omitempty"`
+	// Subkind refines the "task-notification" kind. It is set to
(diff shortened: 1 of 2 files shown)


Roasbeef:

Single omitempty subkind field + typed MessageOriginSubkind const, mirroring the TS task-notification variant. Round-trip and omitempty tests added. CI green; Integration skipped as expected.

PR #173, opened 2026-07-20
hooks: add UserPromptSubmit source and SessionStart fork

Roasbeef:
## What

TS SDK v0.3.215 touches two hook inputs:

- **`UserPromptSubmitHookInput.source`** — new discriminator naming who authored/injected the prompt: `user` (interactive composer), `sdk` (`-p` / Agent SDK), `loop_wakeup` (dynamic /loop), `schedule_wakeup` (scheduled-task fire), `system` (other machine-injected turns). Trial-gated: currently only set for Anthropic-internal sessions, so external payloads omit it — modeled as `omitempty`. Mapped through both the legacy and SDK hook-input builders.
- **`SessionStart` source** — widened with `fork`. `SessionStartInput.Source` is already a plain string mapped in both builders; only the doc comment needed updating.

## Tests

Added `TestGetHookInput_UserPromptSubmit_Source` (present + absent) and `TestGetHookInput_SessionStart_ForkSource`.

### DIFF
--- a/options.go
+++ b/options.go
@@ -1731,6 +1731,14 @@ func (i PostToolUseInput) Base() BaseHookInput { return i.BaseHookInput }
 type UserPromptSubmitInput struct {
 	BaseHookInput
 	Prompt string `json:"prompt"`
+	// Source names who authored or injected the prompt: "user" (interactive
+	// composer), "sdk" (non-interactive entrypoint, -p / Agent SDK),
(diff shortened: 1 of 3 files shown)


Roasbeef:

UserPromptSubmit source threaded through both hook-input builders (the field-mapping path, not plain unmarshal — verified by the failing-then-passing test); SessionStart fork is a doc-only widening on an already-mapped string. omitempty is correct given the trial-gating. Tests cover present/absent source and fork. CI green; Integration skipped as expected.

PR #174, opened 2026-07-20
messages: add subagent fields to tool_progress

Roasbeef:
## What

TS SDK v0.3.215 adds three fields to `SDKToolProgressMessage`:

- **`heartbeat`** — this frame is a liveness heartbeat rather than a content-bearing progress update.
- **`subagent_type`** — the Task-tool subagent driving the tool.
- **`subagent_retry`** — an in-flight subagent retry: `{agent_id, attempt, max_retries, retry_delay_ms, error_status, error_category}`. `error_status` is `number | null` (null for connection errors with no HTTP response), modeled as `*int` to preserve the explicit null.

Adds the three fields on `ToolProgressMessage` plus a `SubagentRetry` struct.

## Tests

Added `TestToolProgressMessageSubagentFieldsRoundTrip` (retry with HTTP status, connection error with null status, and omitempty when absent).

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -499,8 +499,28 @@ type ToolProgressMessage struct {
 	ParentToolUseID    *string `json:"parent_tool_use_id"`   // Parent tool if nested
 	ElapsedTimeSeconds float64 `json:"elapsed_time_seconds"` // Time elapsed
 	TaskID             *string `json:"task_id,omitempty"`    // Task ID if associated with a task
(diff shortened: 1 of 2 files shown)


Roasbeef:

Three additive omitempty fields on ToolProgressMessage plus a SubagentRetry struct mirroring the TS shape; error_status correctly modeled as *int to preserve the explicit null (test exercises both the HTTP-status and null-connection-error paths). CI green; Integration skipped as expected.

PR #175, opened 2026-07-20
options: add v0.3.215 settings fields

Roasbeef:
## What

TS SDK v0.3.215 adds three `Settings` fields:

- **`processWrapper`** — corporate launcher argv prefix for the background-agent supervisor, the sessions/workers it hosts, and other covered background processes. Equivalent to `CLAUDE_CODE_PROCESS_WRAPPER` (env takes precedence). Honored from managed, `--settings`/SDK-supplied, and user settings, in that precedence order; project/local ignored.
- **`feedbackDrafts`** — `'notify'` (default) | `'quiet'` | `'off'`, controlling the model-drafted `SendFeedback` tool (`off` disables it entirely).
- **`vimInsertModeRemaps`** — vim INSERT-mode two-char key remaps to `<Esc>` (e.g. `{"jj": "<Esc>"}`); applies when `editorMode` is `"vim"`.

All `omitempty`, modeled as plain `string` / `map[string]interface{}` to match the surrounding Settings style.

## Tests

Added a round-trip + omitempty subtest for the three fields.

## Deferred (no struct change)

- `effortLevel: 'max'` session-scoped value and the `PartialSettings` conditional typing — `EffortLevel` is already a string alias, so `max` is representable without a change.

### DIFF
--- a/options.go
+++ b/options.go
@@ -421,14 +421,19 @@ type Settings struct {
 	DisableAutoMode                   string                  `json:"disableAutoMode,omitempty"`
 	SSHConfigs                        []SettingsSSHConfig     `json:"sshConfigs,omitempty"`
 	// ClaudeMD is CLAUDE.md-style instructions injected as organization-managed memory. Honored only from managed / policy settings. Mirrors sdk.d.ts v0.3.150 L5343.
-	ClaudeMD              string   `json:"claudeMd,omitempty"`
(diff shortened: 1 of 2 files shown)


Roasbeef:

Three additive omitempty Settings fields modeled as plain string/map to match the surrounding style; vimInsertModeRemaps correctly placed by editorMode. effortLevel:'max' and the PartialSettings typing correctly deferred (already representable). Round-trip + omitempty test added. CI green; Integration skipped as expected.

Issue #189, opened 2026-08-10
Parity with TypeScript Agent SDK v0.3.226


Roasbeef:
Tracking issue for the weekly TypeScript SDK catchup. Prior shipped parity: v0.3.220 (#178). Baseline `9d795db`.

The `sdk.d.ts` delta is 649 lines, but roughly 40% is quote-style churn from an upstream prettier config change and doc rewording. Real surface delta below, one PR per surface.

- [ ] 1. `resumeDropsTurn` option + `--resume-drops-turn` flag
- [ ] 2. `MessageOrigin`: `unclassified` kind + `peer-send-message` subkind
- [ ] 3. `ModelRefusalFallbackMessage.Scope` (`session` / `local`)
- [ ] 4. Sandbox credential structured masking (`mask` files, `extract`, `decode`, `maskClaims`, `maskDuplicates`)
- [ ] 5. Sandbox credential AWS SigV4 re-signing (`awsPairs`, `sigv4`)
- [ ] 6. Settings `dialogExpiry` + `crossSessionInbound`
- [ ] 7. Marketplace `archive` plugin source
- [ ] 8. Docs refresh to v0.3.226

### Deferred

- `OnElicitation` / `OnUserDialog` out-of-band responses — both callbacks now take the envelope `requestId` and may return `null` to suppress the SDK's own control_response. Same transport-contract family as the `canUseTool` null path deferred in v0.3.201; needs an API design call (sentinel error vs. third return value) rather than a mechanical port.
- `bridge.d.ts` classification changes (`CredentialsRejection`, `CreateSessionFailure.reason`) — browser bridge, not the wire protocol this SDK models.
- `sdk-tools.d.ts` churn (`Feedback.failure_mode`/`task_category`, `RemoteTrigger.create_webhook_trigger`) — outside the curated tool-input subset.
- Zod-internal schema plumbing and the single-quote reformatting of the exported string-const arrays.

Roasbeef:
All eight PRs merged. Go SDK is at TypeScript SDK v0.3.226 parity as of `eb502fc`.

- #190 `resumeDropsTurn` fork-point guard (+ `--resume-drops-turn`)
- #191 `MessageOrigin` `unclassified` kind + `peer-send-message` subkind
- #192 `ModelRefusalFallbackMessage.Scope`
- #193 structured sandbox credential masking (`mask` files, `extract`, `decode`, `maskClaims`, `maskDuplicates`)
[...]

PR #190, opened 2026-08-10
options: add resumeDropsTurn fork-point guard

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 1".

A truncating `--resume-session-at` resume drops whatever sits past the fork point with no validation. v0.3.226 pairs it with `--resume-drops-turn` (`sdk.d.ts` L1892), which names the prompt UUID of the turn the resume means to discard; the CLI then checks at fork time that every entry in the discarded range belongs to that turn and refuses the resume otherwise — the case that bites is a queued user message or task notification the session absorbed mid-turn that the caller's view of the conversation had not yet observed.

- `SessionOptions.ResumeDropsTurn` + `WithResumeDropsTurn`, documenting the refusal contract: an `error_during_execution` result prefixed `Resume rejected by --resume-drops-turn:`, deterministic, so consumers route it to their rewind-recovery path instead of retrying.
- The field doc also carries upstream's fork-point rule — fork at the kept turn's last chain entry, whatever it is, since `ResumeSessionAt` accepts any chain UUID and an assistant-UUID fork point deliberately refuses for end-turn tool sessions, `shouldQuery: false` transcript appends, and interrupted turns that completed a tool.
- `transport.go` emits `--resume-drops-turn <uuid>` alongside `--resume-session-at`.
- Unit coverage: `transport_test.go` asserts the flag/value pair is emitted when set and absent when not; `options_test.go` covers the builder.
- Integration: `TestIntegrationResumeDropsTurn` is a skip stub. The installed CLI (2.1.222) has no `--resume-drops-turn` — `--resume-session-at` is in the binary, this is not — so a live run would die on an unknown option rather than exercise the guard. Logged in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:

Checked the flag emission against `sdk.mjs`: the TS SDK pushes `--resume-drops-turn=${value}` immediately after `--resume-session-at`, gated on `!== void 0`. This uses the space-separated form and an empty-string gate, matching the neighboring `--resume-session-at` block in `transport.go` — commander parses both, and repo consistency wins over byte-matching the TS argv.

The skip stub is the honest call: `strings /usr/local/bin/claude` finds `resume-session-at` and not `resume-drops-turn` on 2.1.222, so a live run would exit on an unknown option and the failure would say nothing about the SDK.

CI green on all matrix entries (Lint, Test, 5 Build targets; Integration skipped as usual on PRs).

VERDICT: APPROVE

PR #191, opened 2026-08-10
messages: add unclassified origin kind and peer-send-message subkind

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 2".

- `MessageOriginSubkindPeerSendMessage` (`sdk.d.ts` L4223) — a coordinator co-member SendMessage delivery: model-authored text from another of the same user's sessions, verified by server-stamped receiver co-membership. It lands as a `task-notification` for prompt-authority purposes but stays distinguishable, because the receive-side `crossSessionInbound` setting applies to it and not to a scheduled trigger. Without the subkind a consumer cannot tell the two apart.
- `MessageOriginKindUnclassified` (`sdk.d.ts` L4227) — a delivery the harness could not attribute to any other kind; carries no additional fields.
- The `Subkind` doc no longer claims the field means "scheduled trigger"; it now names both members.
- Unit coverage: round-trip tests for both values in `messages_test.go`, including that `unclassified` carries no `from`/`subkind`.
- No integration test: neither value is triggerable from a local CLI run — both require a live coordinator/cross-session delivery. The existing origin-parity slots stay as-is.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -248,6 +248,10 @@ const (
 	// MessageOriginKindObserverActivity marks an activity digest routed to
 	// the observer. It carries no additional fields.
 	MessageOriginKindObserverActivity MessageOriginKind = "observer-activity"
+	// MessageOriginKindUnclassified marks a delivery the harness could not
+	// attribute to any of the other kinds. It carries no additional fields
+	// (sdk.d.ts v0.3.226 L4227).
+	MessageOriginKindUnclassified MessageOriginKind = "unclassified"
 )
 
 // MessageOrigin describes the originating actor for a message.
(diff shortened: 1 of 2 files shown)


Roasbeef:

Both values verified against `sdk.d.ts` v0.3.226: `subkind?: 'scheduled-trigger' | 'peer-send-message'` at L4223 and the bare `{ kind: 'unclassified' }` arm at L4227. Neither carries extra fields, so the existing `MessageOrigin` shape covers them — no struct surgery needed.

Worth flagging for the settings PR later in this cycle: `peer-send-message` exists precisely so `crossSessionInbound` can key off it. The two land in the same cycle, which is what you want.

VERDICT: APPROVE

PR #192, opened 2026-08-10
messages: add scope to model refusal fallback

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 3".

`SDKModelRefusalFallbackMessage` gained `scope` (`sdk.d.ts` L4265). Until now the message implied one thing: the primary model refused, the turn was retried on the fallback, and the swap stuck for the session. That is no longer the only shape — a subagent, side question (`/btw`), or background fork can fall back locally, with the session model unchanged.

- `ModelRefusalFallbackScope` with `Session` / `Local` consts, and `ModelRefusalFallbackMessage.Scope` (omitempty).
- Older CLIs omit the field and only ever emitted the session-wide form, so an empty `Scope` means `session`. The field doc says so rather than leaving callers to guess.
- The type doc no longer asserts the swap is always session-persistent.
- Unit coverage: `session` and `local` payloads plus an explicit assertion that an absent `scope` stays empty.
- No integration test: `model_refusal_fallback` needs the model to end a turn with `stop_reason: "refusal"` while a fallback model is configured — the same server-side condition that keeps `TestIntegrationModelRefusalNoFallback` skipped.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -1085,16 +1085,36 @@ type APIRetryMessage struct {
 // MessageType implements Message.
 func (m APIRetryMessage) MessageType() string { return "system" }
 
+// ModelRefusalFallbackScope says how far a refusal fallback reached.
+type ModelRefusalFallbackScope string
+
+const (
+	// ModelRefusalFallbackScopeSession means the main thread fell back and the
+	// session model is swapped for good.
+	ModelRefusalFallbackScopeSession ModelRefusalFallbackScope = "session"
+	// ModelRefusalFallbackScopeLocal means a subagent, side question (/btw), or
+	// background fork fell back: only that response came from the fallback
+	// model and the session model is unchanged.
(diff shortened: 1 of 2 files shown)


Roasbeef:

Verified against `sdk.d.ts` L4265 and the surrounding doc: `scope` is optional and absent from older CLIs, which is why this models it as a plain empty-able string type rather than a pointer — an empty `Scope` is exactly the "treat as session" case, and a `*ModelRefusalFallbackScope` would make every caller nil-check to reach the same answer. Consistent with how `Direction` and `MessageOriginSubkind` are modeled.

The absent-field assertion in the older-CLI subtest is the one that matters; it pins the default-inference contract.

VERDICT: APPROVE

PR #193, opened 2026-08-10
options: add structured sandbox credential masking

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 4".

`sandbox.credentials` grew the biggest surface in this release (`sdk.d.ts` L6436–L6500). Credential *files* were `deny`-only — all-or-nothing, so protecting a `.netrc` meant no tool inside the sandbox could read it. They now support `mask`, and both files and env vars gained the knobs that make masking work on a value something still has to parse:

- `Mode` on files widens to `deny | mask`; `mask` shows sandboxed commands a sentinel-substituted copy and the proxy swaps sentinel→real on egress to `InjectHosts`. Degrades to `deny` on macOS and Windows.
- `Extract` — regex whose capture group 1 marks the credential spans; everything else is preserved, so `.netrc` / JSON / YAML still parse. Without it the whole file (or value) becomes one sentinel.
- `OnExtractNoMatch` — `warn` (default, fail-open) / `deny` (fail-closed) / `error` (abort at setup). The env-var doc records upstream's asymmetry: on a mask entry carrying `Decode` the runtime takes the decode path and never consults it, so only `warn` is accepted there.
- `Decode: "jwt"` + `MaskClaims` — verify the value really is a JWT, then swap in a structurally valid fake, optionally masking only named payload claims so claim-reading clients keep working.
- `MaskDuplicates` (files only, matching upstream) and `InjectHosts` on files (env vars already had it).

Unit coverage: file mask with `extract` + `maskDuplicates` + `injectHosts`, JWT decode across both carriers, and a `deny`-entry test asserting every masking knob stays off the wire.

No integration test: `TestIntegrationSettingsSandboxFields` is already a skip stub — these fields are consumed by sandbox-runtime, and the CLI fixture cannot deterministically inject them through managed settings for the SDK to observe.

### DIFF
--- a/options.go
+++ b/options.go
@@ -789,7 +789,9 @@ type SettingsSandbox struct {
 // sandbox.
 type SettingsSandboxCredentials struct {
 	// Files are credential files or directories to protect. "deny" blocks
-	// reads inside the sandbox.
+	// reads inside the sandbox; "mask" substitutes a sentinel inside the
+	// sandbox (whole-file, or per-Extract capture) and injects the real value
+	// at the proxy. On macOS and Windows "mask" degrades to "deny".
 	Files []SettingsSandboxCredentialFile `json:"files,omitempty"`
 	// EnvVars are environment variables to protect. "deny" unsets the
 	// variable for sandboxed commands; "mask" substitutes a sentinel inside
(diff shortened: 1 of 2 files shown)


Roasbeef:

Checked every field against `sdk.d.ts` L6436–L6500. The asymmetries are upstream's, not transcription slips, and both are called out in the field docs rather than smoothed over:

- `maskDuplicates` exists on files only (a bare env-var value has no "outside the matched span" to sweep).
- `onExtractNoMatch` means different things per carrier — on files `deny` degrades the entry to mode `deny`, and is upgraded to `error` under `filesystem.disabled` since read-denies are dropped there; on env vars a mask entry carrying `decode` never consults the field at all, so only `warn` is accepted.

`MaskDuplicates` is `*bool` so an explicit `false` survives the round trip, matching the other tri-state settings booleans in this file.

VERDICT: APPROVE

PR #194, opened 2026-08-10
options: add sandbox credential SigV4 re-signing config

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 5".

Masking an AWS credential breaks SigV4: the client inside the sandbox signs the request with the sentinel, so the proxy has to recompute the signature with the real key on egress. Two new `sandbox.credentials` members configure that (`sdk.d.ts` L6510, L6527).

- `AWSPairs` (`SettingsSandboxCredentialAWSPair`) — explicit groupings of masked env vars into key-id / secret / optional session-token trios, for deployments whose variables aren't named `AWS_ACCESS_KEY_ID` & co. The conventional trio auto-pairs when masked. The type doc records upstream's usability rules: a member only counts when forwarded as a whole-value `mask` entry (an `Extract`/`Decode` entry doesn't qualify — re-signing needs the whole real value), an unusable key-id or secret drops the pair unless it names a conventional AWS variable, in which case it stays as an inert suppressor so auto-pairing remains overridden, and an unusable session token alone still re-signs without `x-amz-security-token`.
- `SigV4` (`SettingsSandboxCredentialSigV4`) — per-shape policy, `deny` (default, fail closed) or `passthrough`, for the three shapes the proxy cannot re-sign: `Streaming` (aws-chunked, per-chunk signatures chain off the seed so re-signing means rewriting the body), `Presigned` (signature lives in the URL), `SigV4A` (ECDSA — no shared-key HMAC to recompute).

Both are managed/user/CLI-settings only; project settings are ignored.

Unit coverage: round-trip with and without `sessionTokenVar`, partial `sigv4` (asserting the unset member stays off the wire), and an omit-both case.

No integration test — same reason as PR #193: `TestIntegrationSettingsSandboxFields` is an existing skip stub because the CLI fixture cannot inject these through managed settings for the SDK to observe.

### DIFF
--- a/options.go
+++ b/options.go
@@ -802,6 +802,58 @@ type SettingsSandboxCredentials struct {
 	// from user, managed/policy, or CLI (--settings) settings — project
 	// settings are ignored (sdk.d.ts v0.3.201).
 	AllowPlaintextInject *bool `json:"allowPlaintextInject,omitempty"`
+	// AWSPairs groups masked env vars into AWS credential pairs for SigV4
+	// re-signing when the variable names are non-standard; the conventional
+	// AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY / AWS_SESSION_TOKEN trio pairs
+	// automatically when masked. Only honored from user, managed/policy, or
+	// CLI (--settings) settings — project settings are ignored (sdk.d.ts
+	// v0.3.226 L6510).
+	AWSPairs []SettingsSandboxCredentialAWSPair `json:"awsPairs,omitempty"`
+	// SigV4 sets the policy for AWS SigV4 request shapes the proxy cannot
+	// re-sign when they reference a masked credential pair. Only honored from
+	// user, managed/policy, or CLI (--settings) settings — project settings are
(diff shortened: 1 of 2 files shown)


Roasbeef:

Verified against `sdk.d.ts` L6510–L6545. Both members are managed/user/CLI-settings only, matching the neighboring `allowPlaintextInject`.

Two things worth having on record because they are counterintuitive and now live in the type docs: an `awsPairs` member only counts when its variable is a *whole-value* mask (an `extract`/`decode` entry cannot re-sign), and an unusable pair naming a conventional AWS variable is kept as an inert suppressor rather than dropped — otherwise removing a broken override would silently re-enable implicit auto-pairing.

Every `sigv4` policy defaults to `deny`, so an omitted member is fail-closed; the test asserts the unset member stays off the wire rather than serializing an empty string.

Branch carries a `git merge origin/main` for the #193 conflict in the same test block — no force-push, both sets of subtests retained.

VERDICT: APPROVE

PR #195, opened 2026-08-10
options: add dialogExpiry and crossSessionInbound settings

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 6".

- `DialogExpiry` (`sdk.d.ts` L6654) — `60s` / `5m` / `10m` / `never`. Caps how long a permission or user dialog forwarded to a remote client stays parked awaiting an answer, and how long a HELD cross-session message waits on approval, before each resolves to its safe no-action default (cancelled, dropped-with-denial). Defaults to 5m, matching the deadline that used to be fixed. Local-only prompts are unaffected; `CLAUDE_CODE_USER_DIALOG_TIMEOUT_MS` overrides it, and it is read from trusted sources only — the field doc says so, since a checked-in repo settings file setting `never` would be a denial-of-service on the host.
- `CrossSessionInbound` (`sdk.d.ts` L6899) — `accept` / `hold` / `refuse` for inbound peer `SendMessage` traffic. This is the other half of the `peer-send-message` subkind from #191: the subkind exists so this setting can apply to those deliveries. Unset means mode parity — auto-deliver only when the sender's permission-mode class matches this session's, hold a mismatched sender, and hold a class-less sender only while this session bypasses prompts. The doc spells that out because "unset" is not "accept".

Unit coverage extends the existing settings-parity table: round-trip plus omit-when-zero for both.

Struct realignment in the diff is `gofmt` widening the field block for the new names, not a functional change.

### DIFF
--- a/options.go
+++ b/options.go
@@ -396,13 +396,24 @@ type Settings struct {
 	// auto-continue with any answers selected so far. Defaults to never —
 	// auto-continue only runs when explicitly set to "60s", "5m", or "10m".
 	// Mirrors sdk.d.ts v0.3.201.
-	AskUserQuestionTimeout string                          `json:"askUserQuestionTimeout,omitempty"`
-	Agent                  string                          `json:"agent,omitempty"`
-	CompanyAnnouncements   []string                        `json:"companyAnnouncements,omitempty"`
-	PluginConfigs          map[string]SettingsPluginConfig `json:"pluginConfigs,omitempty"`
-	Remote                 *SettingsRemote                 `json:"remote,omitempty"`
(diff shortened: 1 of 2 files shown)


Roasbeef:

Both fields verified at `sdk.d.ts` L6654 and L6899. Modeled as plain strings with omitempty, consistent with `askUserQuestionTimeout` and `daemonColdStart` — the enum is CLI-side and an unknown value must survive the round trip rather than being dropped by a typed constant set.

The `crossSessionInbound` unset semantics are the part that could bite: unset is not `accept`, it is permission-mode-class parity between sender and receiver. That is in the field doc.

VERDICT: APPROVE

PR #196, opened 2026-08-10
options: add archive marketplace plugin source

Roasbeef:
Part of #189 (TypeScript SDK v0.3.226 parity). See `memory/catchup-v0.3.226/PLAN.md` §"PR 7".

- `SettingsMarketplaceSourceArchive` (`sdk.d.ts` L5869) — a plugin served as a zip archive over HTTPS. `SettingsMarketplaceSource` is an opaque map, so the wire contract lives in the const's doc: `url` (the plugin root may sit at the top of the archive or one directory deep — a single wrapping directory is stripped) and optional `sha256`, which verifies every download, refuses the install on mismatch, and doubles as version identity when neither `plugin.json` nor the marketplace entry declares a version. The doc keeps upstream's trap: the update signal is the version string, so changing only the digest while a version is declared does not trigger an update.
- `unsupported` sources may now carry an `error` string saying why the source was rejected (L5871).
- Documented the owner-wildcard form on `StrictKnownMarketplaces` / `BlockedMarketplaces`: `{"source":"github","repo":"owner/*"}` matches every repo under that owner, but **only** in those two managed-settings policy lists. In `marketplace add`, `ExtraKnownMarketplaces`, or `known_marketplaces.json` the wildcard is taken literally and fails to clone (L5909, L6112). Worth spelling out — an admin who assumes it works everywhere gets a silently broken registration rather than an error.

Unit coverage: archive source with and without `sha256`, `unsupported` carrying `error`, and the wildcard round-tripping through both policy lists.

### DIFF
--- a/options.go
+++ b/options.go
@@ -357,8 +357,15 @@ type Settings struct {
 	SubagentStatusLine              *SettingsCommand               `json:"subagentStatusLine,omitempty"`
 	EnabledPlugins                  map[string]interface{}         `json:"enabledPlugins,omitempty"`
 	ExtraKnownMarketplaces          map[string]SettingsMarketplace `json:"extraKnownMarketplaces,omitempty"`
-	StrictKnownMarketplaces         []SettingsMarketplaceSource    `json:"strictKnownMarketplaces,omitempty"`
-	BlockedMarketplaces             []SettingsMarketplaceSource    `json:"blockedMarketplaces,omitempty"`
+	// StrictKnownMarketplaces and BlockedMarketplaces are the managed-settings
+	// policy lists, and are the only place a github entry may use the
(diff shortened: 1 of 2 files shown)


Roasbeef:

Checked all four line refs. The archive source is genuinely new (`sdk.d.ts` L5869, present in all three marketplace-entry copies — extraKnownMarketplaces, strictKnownMarketplaces, blockedMarketplaces); `unsupported.error` and the owner-wildcard prose are upstream doc/shape additions on existing surfaces.

No struct change was needed since `SettingsMarketplaceSource` is an opaque map, which is why the field contract lives in the const doc. The wildcard note earns its place: it is scoped to the two policy lists, and getting it wrong elsewhere produces a clone failure rather than a config error.

VERDICT: APPROVE

PR #197, opened 2026-08-10
docs: refresh TypeScript SDK parity to v0.3.226

Roasbeef:
Closes the v0.3.226 cycle (#189). See `memory/catchup-v0.3.226/PLAN.md` §"PR 8".

- README parity section: a v0.3.226 block covering all seven landed PRs, the merged PR list, and the deferred list with reasons.
- `messages.go`: adopts upstream's clarified accounting docs on `ResultMessage` (`sdk.d.ts` L4445–L4460, L4487–L4502). These matter for anyone summing costs — `Usage` is main-agent-loop only and per-turn, while `TotalCostUSD` / `ModelUsage` are cumulative across turns in a streaming-input session, so the correct read is the latest result, not a sum across results. Also records that resumed sessions start fresh and a mid-session `/clear` resets the running total.
- The "four catchup cycles" line was stale several cycles back; it no longer carries a count.

The deferred list calls out one item that is a design decision rather than a skip: `OnElicitation` / `OnUserDialog` now take the envelope `requestId` and may return `null` to suppress the SDK's own `control_response` when the consumer answered out of band. That is the same transport-contract change as the `canUseTool` null path deferred in v0.3.201, and `requestId` alone is useless without the suppression channel — worth deciding deliberately (sentinel error vs. third return value) rather than picking a shape mid-catchup.

Also documented, as follow-up rather than a fix: with no `OnUserDialog` registered the Go SDK answers `request_user_dialog` with `cancelled`, while the TS SDK stays silent so a renderer-bearing client or the worker's park deadline settles it. That divergence predates this release — v0.3.226 only reworded the doc comment to match behavior the TS SDK already had — so changing it belongs in its own PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ stdin/stdout, giving you access to Claude's tool use, extended thinking,
 session management, and hook system.
 
 This repository tracks the official TypeScript Agent SDK surface through the
-v0.3.220 catchup work, using Go idioms where the API shape differs.
+v0.3.226 catchup work, using Go idioms where the API shape differs.
 
 ```mermaid
 flowchart TB
@@ -210,7 +210,7 @@ For detailed guides and examples, see [docs/examples/](docs/examples/):
 ## TypeScript SDK Parity
 
 The SDK tracks the upstream TypeScript Agent SDK release cadence. Coverage
-landed across four catchup cycles.
+landed across a series of catchup cycles, one per upstream release picked up.
 
 The v0.2.119 catchup added:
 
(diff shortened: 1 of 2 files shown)


Roasbeef:

Doc-only apart from the `ResultMessage` comments, which are transcribed from `sdk.d.ts` L4445–L4460 and L4487–L4502. Go has one `ResultMessage` where TS splits success and error, so both upstream copies collapse to a single set of field docs — the text is identical upstream, so nothing is lost.

The parity block covers all seven merged PRs, and the deferred list distinguishes the standing no-Go-surface skips from the one real design decision (the elicitation/dialog out-of-band `null`).

VERDICT: APPROVE

Issue #199, opened 2026-08-17
Parity with TypeScript Agent SDK v0.3.233


Roasbeef:
Weekly catchup against `@anthropic-ai/claude-agent-sdk@0.3.233` (claudeCodeVersion 2.1.233). Prior shipped parity: v0.3.226 (#189).

The `sdk.d.ts` delta looks large at ~700 lines, but most of it is the new `command` plugin source repeated across five copies of the marketplace schema. The one genuinely new surface is `SDKContextUsage` — a structured twin of the `/context` report that rides as a wrapper-level sibling on the synthetic assistant message delivering the markdown table. Everything else is additive settings and init fields. No breaking changes.

### Planned PRs

- [ ] 1 — `context_usage` on the assistant message + `AssistantContextUsage` types (`sdk.d.ts` L3058, L3152, L3218)
- [ ] 2 — `terminal_slash_commands` on the init message (L4712)
- [ ] 3 — `command` plugin source + `disableCommandPluginSources` managed gate (L5974, L6898)
- [ ] 4 — `additionalMarketplaces` / `allowedMarketplaces` settings aliases (L6023, L6468)
- [ ] 5 — `forceLoginGatewayUrl` setting (L6914)
- [ ] 6 — docs refresh to v0.3.233, incl. the `queued_notifications` capability

### Deferred

- `sdk-tools.d.ts` churn (`ReadNotifications`, `ProposeGoal`, `RemoteTrigger.list_runs`/`get_run_log`, `Bash.backgroundEndsWithFinalResponse`, commit `branch`, `thinking_tokens`) — outside the curated `tool_inputs.go` subset, same call as prior cycles.
- `bridge.d.ts` close codes 4093/4094 and the `isCreateSessionFailure` narrowing — browser bridge is alpha with no Go surface.
- `attribution` index signature — Go already ignores unknown keys on decode.

Roasbeef:
All six PRs merged. Go SDK is at parity with TypeScript Agent SDK v0.3.233 (claudeCodeVersion 2.1.233).

- #200 — `context_usage` on the assistant message + `AssistantContextUsage` types
- #201 — `terminal_slash_commands` on the init message
- #202 — `command` plugin source + `disableCommandPluginSources` managed gate
- #203 — `additionalMarketplaces` / `allowedMarketplaces` settings aliases
- #204 — `forceLoginGatewayUrl`
[...]

PR #200, opened 2026-08-17
messages: add context_usage to the assistant message

Roasbeef:
In this PR, we teach `AssistantMessage` about `context_usage`, the structured twin of the `/context` report that TS SDK v0.3.233 added at `sdk.d.ts` L3058. When a user runs `/context`, the CLI answers with a synthetic assistant message carrying a markdown table; from CLI 2.1.233 on it also attaches this field alongside, so a consumer can render the usage card without parsing markdown out of the text block. The markdown stays the canonical fallback for older CLIs.

Like `resumed_from_incomplete_thinking`, it's a wrapper-level sibling and never lands inside `message.content`, so it isn't replayed to the model.

## Why not the existing context-usage type

We already have `SDKControlGetContextUsageResponse` in `query_types.go`, and it would be tempting to reuse it. It's a different wire shape: camelCase, the answer to the `/context` **control request**, and it carries the full internal breakdown (`gridRows`, `systemPromptSections`, `messageBreakdown`, `apiUsage`). The new one is snake_case, rides on the message, and is a deliberately smaller subset that upstream commits to evolving additively (a breaking reshape would ship as a sibling field). So we add `AssistantContextUsage` next to it rather than trying to unify the two, and the doc comment on each says which is which.

The two discriminators (`over_limit.kind`, `categories[].kind`) get named string types with consts instead of bare strings. The TS docs are explicit that rows must be classified on `kind` and never on the display `name`, so that warning is carried into the Go comment where someone reading `Name` will actually see it.

## Testing

Unit coverage in `messages_test.go` parses a full report (over-limit, all four category kinds, MCP tools, memory files, agents, skills with a plugin name), plus a table for the optional-field cases: absent entirely on older CLIs, and present-but-within-limit where `over_limit` and `skills` are both omitted. Round-trip omitempty is asserted on both the message field and the report itself.

`TestIntegrationAssistantContextUsage` drives a real `/context` query and checks the invariants when the field shows up: every row classifies, the `used` rows sum to `total_tokens` (deferred rows are out-of-window tool schemas and excluded from the math), and `over_limit` only appears when the total actually exceeds the window. The CLI installed in CI is 2.1.222, which answers with the table alone, so the test currently skips with that reason after doing the live round trip. It'll start asserting for real the moment the binary catches up, no code change needed.

Part of #199. See `memory/catchup-v0.3.233/PLAN.md` §"PR 1".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2360,3 +2360,80 @@ func TestIntegrationResumeDropsTurn(t *testing.T) {
 	// unattributable entry sits in the discarded range.
 	t.Skip("not triggerable from CLI: installed CLI 2.1.222 does not implement --resume-drops-turn; tracked in INTEGRATION-FOLLOWUPS.md")
 }
+
+// TestIntegrationAssistantContextUsage drives /context against the live CLI and
+// checks the structured sibling on the assistant message that carries the
+// markdown table. CLIs older than 2.1.233 answer /context with the table alone,
+// so the assertions are conditional on the field being attached at all.
+func TestIntegrationAssistantContextUsage(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	client, err := NewClient(isolatedClientOptions(t)...)
+	require.NoError(t, err)
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+
+	var usage *AssistantContextUsage
+	for msg := range client.Query(ctx, "/context") {
+		assistant, ok := msg.(AssistantMessage)
+		if !ok || assistant.ContextUsage == nil {
+			continue
+		}
+		usage = assistant.ContextUsage
+		break
+	}
+
+	if usage == nil {
+		t.Skip("CLI did not attach context_usage to the /context result; " +
+			"needs a CLI at 2.1.233 or newer")
+	}
+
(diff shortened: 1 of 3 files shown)


Roasbeef:
Reviewed the diff against `sdk.d.ts` v0.3.233 L3058/L3152/L3218. Field-by-field the Go shape matches, including the two required-vs-optional distinctions that are easy to get backwards: `over_limit` and `skills` are the only optional members, and both are modeled so they can stay off the wire, while `categories`/`mcp_tools`/`memory_files`/`agents` are required and correctly lack `omitempty`.

Keeping this separate from `SDKControlGetContextUsageResponse` is the right call. They're different wire shapes (snake vs camel) with different producers, and collapsing them would have forced one of the two to carry fields the other never sends.

The integration test earns its keep even while skipping: it does the real `/context` round trip first and only skips after finding no attached field, so it's checking a live condition rather than short-circuiting on a version constant. The `used`-rows-sum-to-total assertion with deferred rows excluded is the invariant most likely to catch a producer-side reshape.

CI: green across all five build matrix entries, Lint, and Test. The first Test run failed on `TestProtocolSendMessage` ("timeout waiting for Initialize to complete"), which is unrelated to this diff (nothing here touches `protocol.go`) and passed on rerun. Same flake family as the `TestProtocolInitializeOptions` context-deadline flake from the v0.3.201 cycle.

VERDICT: APPROVE

PR #201, opened 2026-08-17
messages: add terminal_slash_commands to the init message

Roasbeef:
In this PR, we add `terminal_slash_commands` to `SystemMessage`, from `sdk.d.ts` L4712. It's the subset of `slash_commands` whose UX is bound to the local terminal, `exit` and `statusline` being the examples upstream gives. A phone or other remote UI should hide these from its command menu; desktop surfaces can keep them.

The CLI sends the field only when it's non-empty, so an empty slice on our side means either a CLI that predates the tag or a session where none of the advertised commands carry it. Showing everything is the correct fallback in both of those cases, which is why this stays a plain `omitempty` slice and not a pointer: there's no behavioral difference between "absent" and "empty" worth modeling.

## Testing

`messages_test.go` gets a parse test with a tagged init payload, plus an omitempty round-trip, and the pre-existing older-CLI init test now also asserts the field stays empty when the CLI doesn't send it.

`TestIntegrationInitTerminalSlashCommands` pulls the real init message off a live query and asserts the containment invariant: every terminal command has to also appear in `slash_commands`. The 2.1.222 binary in CI tags nothing, so it skips with that reason after the round trip, and starts asserting once the binary catches up.

Part of #199. See `memory/catchup-v0.3.233/PLAN.md` §"PR 2".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2360,3 +2360,39 @@ func TestIntegrationResumeDropsTurn(t *testing.T) {
 	// unattributable entry sits in the discarded range.
 	t.Skip("not triggerable from CLI: installed CLI 2.1.222 does not implement --resume-drops-turn; tracked in INTEGRATION-FOLLOWUPS.md")
 }
+
+// TestIntegrationInitTerminalSlashCommands checks the terminal-bound subset of
+// the init message's advertised slash commands. CLIs older than 2.1.233 tag
+// nothing, in which case the field is absent and a remote UI showing every
+// command is the correct fallback.
+func TestIntegrationInitTerminalSlashCommands(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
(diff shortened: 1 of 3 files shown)


Roasbeef:
Checked the placement first, since `terminal_slash_commands` at `sdk.d.ts` L4712 sits inside `SDKSystemMessage`, not the initialize control response. It landed on `SystemMessage` next to `SlashCommands`, which is the right home.

The plain-slice-over-pointer choice is correct and the comment explains why: absent and empty both mean "show everything", so there's no state a pointer would let a consumer distinguish that's worth acting on.

The containment assertion (`assert.Subset`) is the useful one here, since a producer bug that tagged a command not present in `slash_commands` would otherwise pass silently. Good that it's asserted in both the unit test and the live path.

CI: green across the matrix, Lint, and Test on the first run.

VERDICT: APPROVE

PR #202, opened 2026-08-17
options: add the command plugin source and its managed gate

Roasbeef:
In this PR, we add the `command` marketplace source (`sdk.d.ts` L5974, repeated across all five copies of the marketplace schema) plus `disableCommandPluginSources`, the managed-settings gate that turns it off (L6898).

Unlike every other source, a `command` plugin isn't fetched, it's produced. The marketplace declares a shell command; that command prints exactly one absolute path on stdout, exits 0, and has to leave a complete plugin sitting at that path. It runs through the platform shell from the user's home directory.

Two details land in the doc comment because they're the ones that bite in practice. The path is re-resolved on every install and update plus once per session in the background, so it can differ run to run. And under `mode: "link"` the cache entry points at the printed directory in place instead of copying it (no size limit, macOS/Linux), which flips the lifetime contract: the directory has to stay valid for as long as Claude Code runs, and a changed printed path becomes the only signal of new content.

Running a marketplace-declared command on the user's machine is a bigger ask than fetching an archive, which is what `disableCommandPluginSources` is for. Left unset it follows `allowManagedHooksOnly`, on the reasoning that an org that locked hook execution down to managed settings didn't mean to leave arbitrary plugin-resolution commands wide open. It stays a `*bool` for exactly that reason: unset and `false` are different states here, so the key has to be able to stay off the wire.

The source itself is one const on `SettingsMarketplaceSourceKind`, same shape as the `archive` source in #196, since `SettingsMarketplaceSource` is an open map and the keys live in the comment.

## Testing

`options_test.go` round-trips the source both fully populated (`timeout`, `mode`) and bare, asserting the optional keys stay off the wire, plus a table on the gate covering `true`, `false`, and nil-omits-the-key.

For the integration side, `TestIntegrationSettingsOptions` gets a `disable_command_plugin_sources` case that pushes the gate through `--managed-settings` and confirms the CLI comes up on it.

The source variant does **not** get a live case, and that's worth flagging: I tried it first and the handshake stalled for the full 120s timeout. Running it against the bare CLI shows why:

```
$ claude --managed-settings '{"extraKnownMarketplaces":{"internal-export":{"source":{"source":"command",...}}}}' -p "say ok"
Contact your administrator.

Detail: parent managed settings: Failed schema validation. This field was ignored.
```

The installed 2.1.222 binary predates the variant, so it fails schema validation on the managed tier and the init never completes. That half is parked in `TestIntegrationSettingsMiscFields` alongside the other source variants the SDK doesn't drive, with the version blocker written down.

Part of #199. See `memory/catchup-v0.3.233/PLAN.md` §"PR 3".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2091,6 +2091,24 @@ exec "$real_cli" "$@"
 		require.NoError(t, json.Unmarshal([]byte(argValue(t, argv, "--managed-settings")), &got))
 		assert.Equal(t, want.Env, got.Env)
 	})
+
+	// Only the gate is asserted live. Feeding a command-sourced marketplace to
+	// a CLI that predates the variant makes it fail schema validation on the
+	// managed tier and stall the handshake, so that half stays in
+	// TestIntegrationSettingsMiscFields with the other source variants until
+	// the installed binary is new enough to accept it.
+	t.Run("disable_command_plugin_sources", func(t *testing.T) {
+		disable := true
+
+		argv := runWithArgvCapture(t, WithManagedSettings(Settings{
+			DisableCommandPluginSources: &disable,
+		}))
+
+		var got Settings
+		require.NoError(t, json.Unmarshal([]byte(argValue(t, argv, "--managed-settings")), &got))
+		require.NotNil(t, got.DisableCommandPluginSources)
+		assert.True(t, *got.DisableCommandPluginSources)
+	})
 }
 
 func TestIntegrationSettingsManagedOrgFields(t *testing.T) {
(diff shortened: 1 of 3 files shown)


Roasbeef:
Verified the const against all five copies of the schema in `sdk.d.ts` v0.3.233 (L5974, L6204, L6429, L6644, L6859) — same variant repeated, so one const covers it, consistent with how `archive` landed in #196.

The `*bool` on the gate is load-bearing and the comment says why: unset defers to `allowManagedHooksOnly` while `false` is an explicit allow, so collapsing to a plain `bool` would silently turn "inherit" into "allow" on every serialization.

The `mode: "link"` lifetime contract is the part of this that will bite someone, and it's documented where they'll see it rather than left to the reader to go find upstream.

Good call cutting the source variant out of the live test instead of leaving a 120s hang in the suite. The bare-CLI repro in the description is the right evidence: the field fails managed-tier schema validation on 2.1.222, so a "passing" test there would only have been proving the CLI ignored it. Parking it in `TestIntegrationSettingsMiscFields` with the version blocker written down is the honest slot, and the gate half still gets a real live assertion through `--managed-settings`.

CI: green across the matrix, Lint, and Test. Mergeable clean against main after #200 and #201.

VERDICT: APPROVE

PR #203, opened 2026-08-17
options: add the marketplace settings alias keys

Roasbeef:
In this PR, we add `additionalMarketplaces` (`sdk.d.ts` L6023) and `allowedMarketplaces` (L6468). Neither is a new capability: the CLI reads them exactly as if they were spelled `extraKnownMarketplaces` and `strictKnownMarketplaces`, so they take the same Go types as the keys they alias, and `allowedMarketplaces` inherits the managed-settings-only restriction from its canonical twin.

The reason this is worth more than two struct fields is the migration hazard, which belongs in the doc comment where someone reaching for the shorter name will actually read it. A client that predates the alias ignores it outright. So a settings file written using only the new spelling reads, to that client, as declaring no marketplaces at all, and its settings sync will happily upload it that way. For `allowedMarketplaces` that failure mode is the dangerous direction: an allowlist that silently isn't there means unrestricted, not blocked. Setting both spellings in one file isn't a fix either, since the alias loses and gets dropped with a warning. The comments say to keep using the canonical spelling until the older clients sharing that file are gone.

We deliberately don't fold the alias into the canonical field on decode. Resolving that conflict is the CLI's job, and merging them here would paper over exactly the misconfiguration the CLI warns about.

## Testing

`options_test.go` covers the serialization contract in both directions, including a case that decodes all four keys side by side to prove the aliases stay independent of the canonical fields, plus the nil-omits-both case.

`TestIntegrationSettingsOptions` gets a `marketplace_aliases` case pushing both keys through `--managed-settings` and asserting they survive the trip under their own spellings without being folded. Worth noting what this does and doesn't prove: 2.1.222 predates the aliases so it won't resolve them, but it does accept them without tripping managed-tier schema validation, which is the part that would break the handshake. I checked that directly rather than assuming, since a nested `command` source in the same position does fail validation and stall init (see #202).

Part of #199. See `memory/catchup-v0.3.233/PLAN.md` §"PR 4".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2109,6 +2109,42 @@ exec "$real_cli" "$@"
 		require.NotNil(t, got.DisableCommandPluginSources)
 		assert.True(t, *got.DisableCommandPluginSources)
 	})
+
+	// Whether the CLI resolves these to their canonical spellings is its
+	// business, and 2.1.222 predates them so it won't. What's ours is that they
+	// serialize under their own keys, don't get folded into the canonical ones,
+	// and don't trip managed-tier schema validation on the way through.
+	t.Run("marketplace_aliases", func(t *testing.T) {
+		want := Settings{
+			AdditionalMarketplaces: map[string]SettingsMarketplace{
+				"vendor": {
+					Source: SettingsMarketplaceSource{
+						"source":  string(SettingsMarketplaceSourceNPM),
+						"package": "@vendor/plugins",
+					},
+				},
+			},
+			AllowedMarketplaces: []SettingsMarketplaceSource{
+				{
+					"source":      string(SettingsMarketplaceSourcePathPattern),
+					"pathPattern": "^/opt/approved/",
+				},
+			},
+		}
+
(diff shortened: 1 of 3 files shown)


Roasbeef:
The types match their canonical twins, which is the whole contract for an alias, and `allowedMarketplaces` correctly inherits the managed-settings-only restriction in its doc comment.

Not folding the alias into the canonical field on decode is the right decision. The CLI's behavior when both appear is to warn and drop the alias, and an SDK that quietly merged them would swallow exactly the misconfiguration that warning exists to surface. The `canonical and alias decode independently` case pins that down so nobody "helpfully" adds the fold later.

The doc comment leads with the failure direction that matters: an allowlist that an older client silently ignores means unrestricted, not blocked. That's the asymmetry someone reaching for the shorter name needs to see.

Good instinct checking CLI acceptance directly rather than assuming it from the #202 result. The two cases genuinely differ (unknown top-level keys pass, an unknown member of the source union doesn't), and assuming either way would have been wrong half the time.

CI: green across the matrix, Lint, and Test.

VERDICT: APPROVE

PR #204, opened 2026-08-17
options: add forceLoginGatewayUrl

Roasbeef:
In this PR, we add `forceLoginGatewayUrl` (`sdk.d.ts` L6914), the Cloud gateway URL that the login flow pre-fills and auto-connects to. It only means anything alongside `forceLoginMethod: "gateway"`, so it sits directly next to that field.

Like the method it pairs with, it's honored only from admin-controlled managed settings (MDM, `managed-settings.json`, policy helper) and ignored in user, project, and remote-delivered settings. That restriction is the interesting part and it's what the doc comment leads with: a login flow that a project-local settings file could point at an attacker-chosen gateway is a credential handoff, so the admin tier is the only one allowed to say where it goes.

## Testing

`options_test.go` round-trips it with `forceLoginMethod` and covers empty-omits-the-key.

`TestIntegrationSettingsOptions` gets a `force_login_gateway_url` case pushing the pair through `--managed-settings`. The login flow itself isn't drivable from an already-authenticated integration session, so what this asserts is the reachable half: the pair survives the managed tier without tripping schema validation and the CLI comes up on it.

Part of #199. See `memory/catchup-v0.3.233/PLAN.md` §"PR 5".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2109,6 +2109,23 @@ exec "$real_cli" "$@"
 		require.NotNil(t, got.DisableCommandPluginSources)
 		assert.True(t, *got.DisableCommandPluginSources)
 	})
+
+	// The login flow itself can't be driven here (the session is already
+	// authenticated), so this covers the reachable half: the pair survives the
+	// managed tier and the CLI starts on it.
+	t.Run("force_login_gateway_url", func(t *testing.T) {
+		want := Settings{
+			ForceLoginMethod:     "gateway",
+			ForceLoginGatewayURL: "https://gateway.corp.example",
+		}
+
+		argv := runWithArgvCapture(t, WithManagedSettings(want))
+		var got Settings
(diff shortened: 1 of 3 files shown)


Roasbeef:
Post-merge review note (the API was throwing 503s while this was being merged, so the verdict landed after the squash rather than before it).

Straightforward addition and it's in the right place, directly next to the `forceLoginMethod` it's only meaningful with.

The doc comment leads with the security reasoning rather than restating the field name, which is the useful thing here: "admin tier only" reads as arbitrary bureaucracy until you notice that a project-local settings file pointing the login flow at an attacker's gateway is a credential handoff.

The integration case is scoped honestly. An already-authenticated session can't drive a login flow, so it asserts the half that is actually reachable (the pair survives the managed tier, CLI starts on it) and the comment says that's what it's doing instead of implying broader coverage.

CI was green across the matrix, Lint, and Test after the merge from main. The conflict with #203 was resolved by merging origin/main rather than rebasing, so the branch history stayed intact.

VERDICT: APPROVE

PR #205, opened 2026-08-17
docs: refresh TypeScript SDK parity to v0.3.233

### LINKED ISSUE #199: Parity with TypeScript Agent SDK v0.3.233

Roasbeef:
In this PR, we close out the v0.3.233 cycle: the README parity section gets the new entries, the version reference at the top moves off v0.3.226, and the deferral list records what we skipped and why.

One code touch, and it's a comment. `queued_notifications` joins the documented values on `SystemMessage.Capabilities`: the CLI advertising it accepts inbound `queued_notification` stream messages and drains them through the `ReadNotifications` tool. Nothing to add for it, since `Capabilities` is already an open `[]string` (that's the point of an open set) and the drain tool lives outside the curated `tool_inputs.go` subset, same as the rest of the `sdk-tools.d.ts` churn.

We also ignore `scratch/`. Every catchup pulls a vendored TS SDK tree in there to diff against, and it's been sitting untracked across enough cycles that a stray `git add -A` would have committed a couple megabytes of unpacked npm tarball. I nearly did exactly that earlier in this cycle.

## What the cycle covered

The `sdk.d.ts` delta read as ~700 lines, but most of it was the `command` plugin source repeated across five copies of the marketplace schema. The one genuinely new surface was `SDKContextUsage`. Everything else was additive settings and init fields, and nothing broke.

Merged: #200 assistant `context_usage`, #201 `terminal_slash_commands`, #202 command plugin source + managed gate, #203 marketplace settings aliases, #204 `forceLoginGatewayUrl`.

Deferred, with reasons written into the README rather than left implicit: the `sdk-tools.d.ts` churn (`ReadNotifications`, `ProposeGoal`, `RemoteTrigger.list_runs`/`get_run_log`, `Bash.backgroundEndsWithFinalResponse`, commit `branch`, `thinking_tokens`), the `bridge.d.ts` close codes 4093/4094 and the `isCreateSessionFailure` narrowing, the `attribution` index signature, and a `ripgrep` settings doc reword.

Closes #199.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 package/*
 .sessions/
 .tasks/
+
+# Vendored TS SDK trees pulled for parity diffing during catchup cycles.
+scratch/
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ stdin/stdout, giving you access to Claude's tool use, extended thinking,
 session management, and hook system.
 
 This repository tracks the official TypeScript Agent SDK surface through the
-v0.3.226 catchup work, using Go idioms where the API shape differs.
+v0.3.233 catchup work, using Go idioms where the API shape differs.
 
 ```mermaid
 flowchart TB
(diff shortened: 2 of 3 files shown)


Roasbeef:
Checked the parity section against the five merged PRs and it matches, including the distinctions that actually matter to someone porting: that `ContextUsage` is a different wire shape from `SDKControlGetContextUsageResponse` rather than a rename, that rows classify on kind and not display name, and that the SDK deliberately doesn't fold the marketplace aliases.

`queued_notifications` as doc-only is right. The field is an open `[]string` by design, so adding a const would imply a closed set that upstream explicitly says isn't one.

The `scratch/` ignore is a good catch and belongs in this PR rather than its own. Untracked-but-not-ignored is the worst of both states, and a few megabytes of unpacked npm tarball landing in a parity commit would be an annoying thing to unwind.

Deferrals are recorded with reasons instead of just listed, which is what makes the next cycle's diff cheap to triage.

CI: green across the matrix, Lint, and Test.

VERDICT: APPROVE

Issue #206, opened 2026-08-24
Parity with TypeScript Agent SDK v0.3.241


Roasbeef:
Tracking issue for the weekly TypeScript SDK catchup.

**TS SDK:** 0.3.241 (npm, 2026-08-24)
**Prior Go shipped:** 0.3.233 (#200–#205)
**Main baseline:** `b36a31a`

The `sdk.d.ts` diff runs 815 lines but only 69 are non-comment — the bulk is doc
expansion on types already modeled here. `sdk.mjs` emits an identical `--flag`
set, so no transport work this cycle. Everything landing is additive. The two
removals upstream (`ExitReason.bypass_permissions_disabled`, and
`SDKControlGetPlanRequest`/`SDKControlGetWorkspaceDiffRequest` deleted outright)
touch surfaces the Go SDK never modeled.

### Planned PRs

- [ ] 01 `init-effort` — `effort` on the init system message, as a tri-state (absent / explicit null / value)
- [ ] 02 `task-started-backgrounding` — `is_backgrounded` + `spawn_depth` on `task_started`
- [ ] 03 `origin-peer-from-mode` — peer origin `fromMode`, plus the `projects-relay` subkind
- [ ] 04 `api-key-source-and-account-on-hold` — `account_on_hold` error code + `ApiKeySource` constants
- [ ] 05 `post-tool-use-classifier-context` — `classifierContext` on PostToolUse hook output
- [ ] 06 `prompt-hook-parity` — `suppressOriginalPrompt` on UserPromptExpansion, `poll_event` prompt source
- [ ] 07 `init-hooks-applied` — `hooks_applied` on the initialize control response
- [ ] 08 `settings-parity-0.3.241` — `syncClaudeAiSkills`, `worktree.location`, `keybindingFlavor`, `autoContinueAtUsageLimit`, `modelSettings`, `spellcheck`
- [ ] 09 `marketplace-headers-helper` — `headersHelper` on marketplace url sources and inline catalog entries
- [ ] 10 `docs-refresh-0.3.241` — parity version bump

### Deferred

- `sdk-tools.d.ts` churn (185 diff lines) — the Go SDK does not mirror tool input schemas.
- `bridge.d.ts` — `onStopTask` callback and the terminal-reason split (`request_rejected` gains `status`/`source`, `malformed_response` becomes its own variant). Deferred in v0.3.220 and v0.3.226 for the same reason: there is no bridge transport here.

[...]

PR #207, opened 2026-08-24
messages: add effort to the init system message

Roasbeef:

`sdk.d.ts` v0.3.241 L4816 adds `effort` to `SDKSystemMessage`:

```ts
effort?: ('low' | 'medium' | 'high' | 'xhigh' | 'max') | null;
```

It reports the effort level the session will send on its next request —
after env overrides, session state, org caps and model-support downgrades.
Same value `get_settings` reports as `applied.effort`.

### The tri-state

This field is not a plain optional. The upstream doc distinguishes two
different "no level" cases, and they mean opposite things to a consumer:

- **explicit `null`** — no effort parameter will be sent at all. A model
  without effort, `CLAUDE_CODE_EFFORT_LEVEL=unset`, or an internal numeric
  budget. This is a definite answer.
- **absent** — the host does not publish the field, or the CLI predates it.
  The applied effort is simply unknown.

The obvious `*EffortLevel` collapses both to `nil`. Worse, so does
`*NullableEffortLevel`: `encoding/json` resolves a `null` into a nil pointer
without ever calling `UnmarshalJSON`, so a pointer field structurally cannot
carry the distinction no matter what the pointee implements.

So it lands as an `omitzero` value type that records presence separately:

```go
type NullableEffortLevel struct {
	Present bool
	Level   *EffortLevel
}
```

`Present` false is absent; `Present` with a nil `Level` is the explicit null;
a non-nil `Level` is the level. Round-trips in all three states, which the
test asserts directly rather than trusting the tag.

Re-emitted inits carry the current value, so the newest frame wins — noted in
the doc comment since a consumer caching the first init would otherwise go
stale.

### Testing

Table-driven parse test over all three states plus `max` (which the init
message accepts but `Settings.modelSettings` does not), and a round-trip test
asserting the marshalled key is present/absent/`null` as expected.

Integration test is a documented skip. Probed against the live CLI 2.1.222:
the init message carries `apiKeySource` and `capabilities` but no `effort`
key. The field is published on Remote Control bridge init frames, and a
subprocess transport is not one, so a live assertion could only re-assert the
absent state.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -740,6 +740,17 @@ func TestIntegrationHookSuppressOriginalPrompt(t *testing.T) {
 	t.Skip("not directly assertable from CLI: suppressOriginalPrompt is a CLI block-message rendering behavior, not surfaced on the SDK transport")
 }
 
+func TestIntegrationInitEffort(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// effort is published on Remote Control bridge init frames; a subprocess
+	// transport is not one. Verified against CLI 2.1.222: the init message
+	// carries apiKeySource and capabilities but no effort key at all, so a
+	// live assertion here could only ever re-assert the absent state.
+	t.Skip("not triggerable from CLI: effort is published on Remote Control bridge init frames, not on the subprocess transport's init message")
+}
+
 func TestIntegrationSessionStartReloadSkills(t *testing.T) {
 	skipIfNoToken(t)
 	skipIfNoCLI(t)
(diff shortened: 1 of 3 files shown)


Roasbeef:

The tri-state modeling is the right call and the reasoning holds: `encoding/json` resolves a `null` into a nil pointer without consulting the pointee's `UnmarshalJSON`, so `*NullableEffortLevel` would have collapsed explicit-null into absent no matter what the type implemented. `omitzero` on a value field is the only shape that survives all three states, and the round-trip test asserts that rather than assuming it.

Checked: `go.mod` declares `go 1.24.0`, which is exactly where `omitzero` landed in `encoding/json` — the tag is not silently ignored on any toolchain able to build this module.

Skipped integration test is justified with a live probe rather than a guess (CLI 2.1.222 emits `apiKeySource` and `capabilities` on init, no `effort` key).

Build, Test and Lint green across all five matrix entries.

VERDICT: APPROVE

PR #208, opened 2026-08-24
messages: add backgrounding and spawn depth to task_started

Roasbeef:

`sdk.d.ts` v0.3.241 adds two fields to `SDKTaskStartedMessage`:

- **L4883 `is_backgrounded?: boolean`** — whether the task was registered in
  the background (true) or in the foreground with the spawning tool call
  blocking on it (false). A resumed subagent is always registered in the
  background. Set for `local_agent` and `local_bash` tasks.
- **L4887 `spawn_depth?: number`** — nesting depth of a spawned subagent
  (`local_agent`) task: 1 for a top-level spawn, N+1 when spawned from inside
  a depth-N agent. Not set on other task types.

Both land as pointers. `false` and `0` are not "absent" here — a foreground
spawn genuinely reports `is_backgrounded: false`, and collapsing that into the
zero value would make it indistinguishable from a `local_workflow` task that
never carries the field at all.

Worth noting for consumers, and called out in the doc comment: `task_started`
reports the state at registration only. A task that *later* moves to the
background does not restate it — that arrives as `patch.is_backgrounded` on a
`task_updated` message, which `TaskUpdatePatch` already models. Reading only
`task_started` and caching the result gets you a stale answer.

### Testing

Two new cases on the existing `TestParseMessageTaskStarted` table: a foreground
depth-1 spawn (asserting `false` survives as a set value, not a nil) and a
nested background depth-2 spawn. The minimum-required-fields case now also
asserts both fields stay nil when absent.

Integration test is a documented skip. Probed against the live CLI 2.1.222 with
a prompt that spawns one Explore subagent — the `task_started` frame carries
`task_id`, `tool_use_id`, `description`, `subagent_type` and `task_type:
"local_agent"`, but neither new field. The runner CLI predates both, so a live
assertion could only re-assert the absent state.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -740,6 +740,18 @@ func TestIntegrationHookSuppressOriginalPrompt(t *testing.T) {
 	t.Skip("not directly assertable from CLI: suppressOriginalPrompt is a CLI block-message rendering behavior, not surfaced on the SDK transport")
 }
 
+func TestIntegrationTaskStartedBackgrounding(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// Probed against CLI 2.1.222 with a prompt that spawns one Explore
+	// subagent: the task_started frame carries task_id, tool_use_id,
+	// description, subagent_type and task_type "local_agent", but neither
+	// is_backgrounded nor spawn_depth. The runner CLI predates both fields,
+	// so a live assertion could only re-assert the absent state.
+	t.Skip("not triggerable from CLI: runner CLI 2.1.222 omits is_backgrounded and spawn_depth from task_started")
+}
+
 func TestIntegrationSessionStartReloadSkills(t *testing.T) {
 	skipIfNoToken(t)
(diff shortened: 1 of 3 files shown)


Roasbeef:

Both fields correctly land as pointers. `is_backgrounded: false` is a real state (the spawning tool call is blocking on the task), not an absent field, and `spawn_depth` has the same problem at 0 — collapsing either into the zero value would make a `local_agent` spawn indistinguishable from a `local_workflow` task that never carries the field.

The doc comment's cross-reference to `TaskUpdatePatch.IsBackgrounded` is the part worth keeping: `task_started` reports registration state only, so a consumer that reads it and caches gets a stale answer once the task moves to the background. That transition already had a home on the patch type; now both halves point at each other.

Skip is backed by a live probe — a real Explore spawn against CLI 2.1.222 produced a `task_started` frame with `subagent_type` and `task_type: local_agent` but neither new field.

CI green.

VERDICT: APPROVE

PR #209, opened 2026-08-24
messages: add peer fromMode and the projects-relay subkind

Roasbeef:

Two additions to `MessageOrigin`.

### `fromMode` on the peer variant (`sdk.d.ts` L4350)

```ts
fromMode?: 'bypass' | 'prompting';
```

The **sending** session's permission class, as declared by the host that
injects the message on local stdin — `bypass` for sessions that run tools
without asking, `prompting` otherwise. A recipient that itself runs without
asking uses this to deliver a same-class message immediately, while a
cross-class or undeclared sender is still held.

That last clause is the part worth reading carefully, and the doc comment
says so explicitly: an **absent** `fromMode` means *hold*, not "assume
prompting". Treating the zero value as a default class would turn a missing
declaration into an immediate delivery, which is exactly backwards. It is
honored only from the injecting host on local stdin.

Lands as a named `PeerFromMode` with two consts rather than a bare string,
per the repo convention for TS union types.

### `projects-relay` subkind (`sdk.d.ts` L4377)

A Claude Code Projects delivery that Anthropic servers composed for the
project's coordinator session and addressed to one of its thread sessions —
the thread's first message, or a relay carrying project messages. Stamped
from server-asserted provenance.

Note the conditional framing, carried into the comment: the harness frames it
as a message from the coordinator session only when it carries the server's
relay stamps. One without them keeps the generic background-notification
frame. The subkind alone does not tell you which you got.

The `Subkind` field doc is updated to list all three refinements instead of
two.

### Testing

The existing peer-origin test gains `fromMode` and asserts it does *not*
marshal when unset (the omitempty behavior matters more than usual here,
given the hold semantics). Two new tests cover the `prompting` class and a
`projects-relay` task-notification origin.

No integration test: neither field is reachable from a subprocess transport.
`fromMode` is set only by a host injecting peer messages on local stdin, and
`projects-relay` requires a server-composed Projects delivery.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -411,6 +411,15 @@ type MessageOrigin struct {
 	Server string            `json:"server,omitempty"`
 	From   string            `json:"from,omitempty"`
 	Name   string            `json:"name,omitempty"`
+	// FromMode is the SENDING session's permission class, as declared by the
+	// host that injects this message on local stdin. It lets a recipient
+	// deliver a same-class message immediately while a cross-class or
+	// undeclared sender is still held at a recipient that runs without
+	// asking — so an empty FromMode must be treated as "hold", not as a
+	// default. Honored only from the injecting host on local stdin, and set
+	// only for the "peer" kind; absent when the host does not declare it
+	// (sdk.d.ts v0.3.241 L4350).
+	FromMode PeerFromMode `json:"fromMode,omitempty"`
 	// SenderTaskID is the observer's task id; set only for the
 	// "observer" kind (sdk.d.ts v0.3.201).
 	SenderTaskID string `json:"senderTaskId,omitempty"`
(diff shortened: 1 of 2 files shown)


Roasbeef:

`FromMode` correctly gets a named type rather than a bare string — it is a closed two-member union upstream, and the repo convention already covers that case.

The important detail is captured in the doc comment and enforced by a test: an absent `fromMode` means *hold*, not "assume prompting". A recipient running without asking delivers same-class messages immediately and holds cross-class or undeclared senders, so treating the zero value as a default class would invert the intended behavior. The test asserting the field does not marshal when unset is load-bearing here, not boilerplate.

`projects-relay` is documented with its conditional framing intact: the harness presents it as a coordinator message only when the server's relay stamps are present, and the subkind alone does not tell a consumer which they got.

`Subkind` field doc updated to list all three refinements.

CI green.

VERDICT: APPROVE

PR #210, opened 2026-08-24
options: add account_on_hold and apiKeySource constants

Roasbeef:

### `account_on_hold`

`sdk.d.ts` L159 adds `'account_on_hold'` to `SDKAssistantMessageError`. One
const, slotted next to the other auth/billing codes.

### `apiKeySource`

`sdk.d.ts` L124 rewrote `ApiKeySource` — four real members, and a doc
explaining that the five it previously listed are legacy:

```ts
export declare type ApiKeySource = 'ANTHROPIC_API_KEY' | 'apiKeyHelper'
  | '/login managed key' | 'none'
  | 'user' | 'project' | 'org' | 'temporary' | 'oauth';
```

Upstream is explicit that `user`/`project`/`org`/`temporary`/`oauth` are kept
only so the type stays backward compatible and **current CLIs never emit
them**.

The Go field is a plain `string` on both `SystemMessage` and `AccountInfo`.
It stays that way — this is an open set the CLI can extend, and retyping a
public field would break consumers for no benefit. So this adds constants for
the four live values instead, with a doc saying the set is not exhaustive.
The legacy five get no constants: there is nothing to branch on.

Naming the values is worth more here than usual because two of them are not
guessable from a Go identifier — `"/login managed key"` has spaces and a
slash, and `"apiKeyHelper"` is lowerCamel where the identifier is not.

Confirmed live rather than assumed: CLI 2.1.222 under a claude.ai OAuth login
emits `"apiKeySource": "none"` on the init message. That is the
subscription-auth case, not an error state — called out in the const doc,
since `none` reads like a failure at a glance.

### Testing

Constants pinned against their wire literals (the point of the test is the
values, not the names), a parse test for the `none` init case, and the
`account_on_hold` code.

### DIFF
--- a/options.go
+++ b/options.go
@@ -2320,14 +2320,44 @@ type AssistantMessageError string
 const (
 	AssistantMessageErrorAuthenticationFailed AssistantMessageError = "authentication_failed"
 	AssistantMessageErrorOAuthOrgNotAllowed   AssistantMessageError = "oauth_org_not_allowed"
-	AssistantMessageErrorBillingError         AssistantMessageError = "billing_error"
-	AssistantMessageErrorRateLimit            AssistantMessageError = "rate_limit"
-	AssistantMessageErrorOverloaded           AssistantMessageError = "overloaded"
-	AssistantMessageErrorInvalidRequest       AssistantMessageError = "invalid_request"
-	AssistantMessageErrorModelNotFound        AssistantMessageError = "model_not_found"
-	AssistantMessageErrorServerError          AssistantMessageError = "server_error"
-	AssistantMessageErrorUnknown              AssistantMessageError = "unknown"
-	AssistantMessageErrorMaxOutputTokens      AssistantMessageError = "max_output_tokens"
+	// AssistantMessageErrorAccountOnHold marks a turn refused because the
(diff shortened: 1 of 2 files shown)


Roasbeef:

Right call leaving `APIKeySource` a plain `string` on both `SystemMessage` and `AccountInfo`. It is an open set the CLI can extend, and retyping a public field would be a breaking change buying nothing — constants give callers the same safety without the churn. Declining to add constants for the five legacy members is also correct: upstream states current CLIs never emit them, so a constant would only invite dead branches.

Naming the values earns its keep given `"/login managed key"` and `"apiKeyHelper"` are not derivable from a Go identifier, and the test pins the wire literals rather than the names.

The `"none"` value is confirmed live rather than assumed — CLI 2.1.222 under claude.ai OAuth emits it on init — and the comment flags that it is the subscription-auth case, not a failure, which is how it reads at a glance.

gosec G101 flagged three of the four as hardcoded credentials. Annotated per-line after confirming a block-level `#nosec` does not propagate to individual const specs. Verified clean with golangci-lint v2 locally before the final push.

CI green.

VERDICT: APPROVE

PR #211, opened 2026-08-24
protocol: add classifierContext to PostToolUse hook output

Roasbeef:

`sdk.d.ts` v0.3.241 L2339 adds `classifierContext?: string` to
`PostToolUseHookSpecificOutput` — host-asserted context shown to the auto-mode
permission classifier alongside a tool call's result.

Lands as `HookJSONOutput.ClassifierContext`, emitted through `buildHookResponse`
gated to `PostToolUse`, following the shape the other typed hook-output fields
already use.

### Why the doc comment is long

This field hands text to the permission classifier *with host-application
framing*. The classifier may weigh a user statement relayed here as user
intent — it can satisfy a consent bar a user turn would satisfy. That makes it
the one hook-output field where getting it wrong has a security consequence
rather than a cosmetic one, so the constraints are carried into the comment
rather than left in the TS source:

- **2000 UTF-16 code units**, a budget shared across every hook contributing
  to one call. Astral characters count as two.
- **Synchronous responses only.** An async hook's late response arrives after
  the result message is frozen and the field is ignored.
- **Only for calls the classifier transcript shows.** Read-only lookups the
  transcript omits, inner REPL calls and remote-engine shells produce no
  per-result line, so context attached to them is silently unused.
- **Rewrite integrity.** Pair an assertion with the rewrite it describes in
  the *same* hook result, so it drops automatically if that rewrite is
  rejected or superseded. But do not return an identity rewrite just to
  achieve that pairing — hooks run in parallel on the original output, so an
  identity rewrite competes last-write-wins with sibling rewrites and can
  clobber a real redaction.

Every one of those fails silently rather than erroring, which is exactly the
kind of thing worth reading at the call site.

### Testing

Unit tests cover emission on PostToolUse, empty-string omission, composition
with an explicit `HookSpecificOutput` map, and riding alongside an
`UpdatedToolOutput` rewrite in one result (the paired-assertion case the
upstream contract calls for). The drop test loops over `PreToolUse`,
`PostToolUseFailure`, `UserPromptSubmit` and `Stop` rather than checking one,
since the leak is the failure mode that matters.

Integration test is a documented skip: the observable effect is an auto-mode
classifier decision, which the SDK transport does not surface — same reason
`TestIntegrationPostToolUseUpdatedToolOutput` is skipped. Asserting only that
the key reached the wire would restate the unit test.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -771,6 +771,17 @@ func TestIntegrationPostToolUseUpdatedToolOutput(t *testing.T) {
 	t.Skip("not triggerable from CLI without a deterministic tool call whose output a PostToolUse hook rewrites; tracked in INTEGRATION-FOLLOWUPS.md")
 }
 
+func TestIntegrationPostToolUseClassifierContext(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// The observable effect is a permission-classifier decision in auto mode,
+	// which the SDK transport does not surface — the same reason
+	// TestIntegrationPostToolUseUpdatedToolOutput above is skipped. Asserting
+	// only that the SDK put the key on the wire would restate the unit test.
+	t.Skip("not directly assertable from CLI: classifierContext only affects auto-mode permission classifier decisions, which are not surfaced on the SDK transport")
+}
+
 func TestIntegrationStopHookBackgroundTasks(t *testing.T) {
 	skipIfNoToken(t)
 	skipIfNoCLI(t)
--- a/options.go
+++ b/options.go
@@ -2602,6 +2602,49 @@ type HookResult struct {
 	// updatedMCPToolOutput field.
 	UpdatedToolOutput interface{}
 
+	// ClassifierContext is host-asserted context shown to the auto-mode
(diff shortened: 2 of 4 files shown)


Roasbeef:

Gating to `PostToolUse` matches the TS envelope and is the right default for this field specifically. Unlike `additionalContext`, which several envelopes accept, `classifierContext` reaches the permission classifier carrying host-application framing — leaking it onto an envelope that never declared it is a security-relevant mistake, not a cosmetic one. The drop test looping four hook types rather than checking one is proportionate to that.

The doc comment is long, and justified. Every constraint it lists fails *silently*: over the 2000 UTF-16 budget, on an async response arriving after the result is frozen, or on a call the classifier transcript omits. A caller has no runtime signal for any of them, so the comment is the only place that information can live.

The rewrite-integrity paragraph is the one I checked hardest against the TS source, since the guidance is counterintuitive — pair an assertion with its rewrite in the same result, but never manufacture an identity rewrite to do so, because hooks run in parallel on the original output and it would compete last-write-wins with a real redaction. Faithful to upstream, and the test covers the legitimate paired case.

CI green.

VERDICT: APPROVE

PR #212, opened 2026-08-24
protocol: honor suppressOriginalPrompt on UserPromptExpansion

Roasbeef:

Two prompt-hook items from this cycle.

### `suppressOriginalPrompt` on UserPromptExpansion (`sdk.d.ts` L8219)

`UserPromptExpansionHookSpecificOutput` now declares the field, with the same
semantics it has on `UserPromptSubmit`: when the decision is `block`, omit the
original prompt from the block message.

`buildHookResponse` hardcoded `HookTypeUserPromptSubmit` in two places — the
gate *and* the `hookEventName` it synthesized. Widening only the gate would
have emitted an expansion hook's output under a `UserPromptSubmit` event name,
so both move: the gate becomes `isSuppressOriginalPromptHook`, sitting next to
the existing `isAdditionalContextHook`, and the envelope now names the hook
that actually produced it.

### `poll_event` prompt source (`sdk.d.ts` L8228)

`UserPromptSubmitHookInput.source` gains `'poll_event'` — the poll-event
channel enqueue-time pass.

This one is **doc-only**. Both `protocol.go` hook-input builders already map
`source` through `getString`, so the value parses today; there is no plumbing
to add. What was missing is the meaning, and it is worth stating: `poll_event`
is the one source where a blocking verdict rejects something *not yet
delivered*. The hook fires when the host submits the event, before its
delivery ack exists. A hook author who assumes a blocking verdict merely
suppresses display will get that backwards.

### Testing

Three subtests on the expansion path: emission under the correct event name
(the regression the two-place hardcode invited), explicit `false` surviving to
the wire, and `UserPromptSubmit` still naming itself.

No integration test. `TestIntegrationUserPromptSubmitSuppressOriginalPrompt`
is already a documented skip — the omission is a CLI block-message rendering
behavior, not something the SDK transport surfaces — and the expansion variant
is unobservable for exactly the same reason.

### DIFF
--- a/options.go
+++ b/options.go
@@ -2005,10 +2005,17 @@ type UserPromptSubmitInput struct {
 	// Source names who authored or injected the prompt: "user" (interactive
 	// composer), "sdk" (non-interactive entrypoint, -p / Agent SDK),
 	// "loop_wakeup" (dynamic /loop wakeup), "schedule_wakeup" (scheduled-task
-	// fire), or "system" (other machine-injected turns: peer/channel messages,
-	// task notifications, auto-continuation). Empty when absent — as of
-	// v0.3.215 it is only set for Anthropic-internal sessions while the field
-	// is trialed; external payloads omit it (sdk.d.ts v0.3.215).
+	// fire), "system" (other machine-injected turns: peer/channel messages,
+	// task notifications, auto-continuation), or "poll_event" (the poll-event
+	// channel enqueue-time pass, added in sdk.d.ts v0.3.241 L8228).
+	//
+	// A "poll_event" pass is the one source where a blocking verdict rejects
+	// something that has not been delivered yet: the hook fires when the host
+	// submits the event, before its delivery ack exists.
+	//
+	// Empty when absent — as of v0.3.215 it is only set for
+	// Anthropic-internal sessions while the field is trialed; external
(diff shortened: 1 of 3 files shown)


Roasbeef:

The catch here is that `buildHookResponse` hardcoded `UserPromptSubmit` in *two* places — the gate and the synthesized `hookEventName`. Widening only the gate would have shipped an expansion hook's output labelled as a submit hook, a wire-correctness bug the unit test would not catch unless it asserted the event name. It does, and both places moved together.

Extracting `isSuppressOriginalPromptHook` alongside the existing `isAdditionalContextHook` keeps the per-field envelope allowlists in one recognizable shape rather than growing another inline condition.

The `poll_event` half is correctly identified as doc-only. Verified both `protocol.go` hook-input builders already map `source` through `getString` — this repo has bitten before on fields added to one builder and not the other, so that was worth confirming rather than assuming.

Rebased onto #211, which landed a `classifierContext` block immediately above this gate and an EOF test block. Resolved by keeping both; confirmed the pre-existing `ignored_on_non-UserPromptSubmit_hooks` subtest still passes, since `PreToolUse` remains outside the widened allowlist.

CI green.

VERDICT: APPROVE

PR #213, opened 2026-08-24
query: add hooks_applied to the initialize response

Roasbeef:

`sdk.d.ts` v0.3.241 L3752 adds `hooks_applied?: boolean` to the initialize
control response.

- **true** — a session's first initialize, or a repeated initialize from the
  process that owns the CLI's stdin. Its hook set replaces the one registered
  earlier.
- **false** — a repeated initialize's hooks were ignored. This is what a
  client joining a remote session *another* client configured gets.
- **absent** — the request carried no hooks, or the CLI predates the field.

### Why it is a pointer

Nil is not false, and conflating them is the mistake worth preventing: a CLI
that predates the field ignored `hooks` on every repeated initialize *without
saying so*. So a nil after a `Reinitialize` that did carry hooks is not
evidence they took effect — it is the absence of evidence either way. The doc
comment says that outright.

`Reinitialize` is where a false is actionable, so the comment points at it. A
caller reattaching to a session it did not configure keeps whatever hooks the
owning client registered, and its own callbacks silently never fire. That is a
failure mode with no other signal.

### Testing

Table-driven unmarshal test over true / false / absent, plus an omitempty
check.

The integration test is a **live** one rather than a hard skip:
`TestIntegrationReinitializeHooksApplied` registers a PreToolUse hook, opens a
stream, and reinitializes. This SDK owns the CLI's stdin, so if the field is
reported at all it must be true. The assertion is conditional on presence, so
it skips cleanly on CLIs that omit it and starts asserting the moment the
runner catches up — same pattern `TestIntegrationAssistantContextUsage` uses.

Verified against the live CLI 2.1.222:

```
=== RUN   TestIntegrationReinitializeHooksApplied
    integration_test.go:2457: CLI does not report hooks_applied on the initialize response
--- SKIP: TestIntegrationReinitializeHooksApplied (1.30s)
```

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2414,6 +2414,53 @@ func TestIntegrationReinitialize(t *testing.T) {
 	require.NotNil(t, resp, "expected a fresh initialize response")
 }
 
+// TestIntegrationReinitializeHooksApplied checks hooks_applied on a repeated
+// initialize. This SDK owns the CLI's stdin, so if the CLI reports the field at
+// all it must report true — the repeated initialize's hook set replaces the one
+// registered earlier. CLIs predating the field omit it, so the assertion is
+// conditional on presence.
+func TestIntegrationReinitializeHooksApplied(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("You are a helpful assistant. Be very brief."),
+		WithMaxTurns(1),
+		WithHooks(map[HookType][]HookConfig{
+			HookTypePreToolUse: {
+				{Matcher: "*", Callback: func(
+					ctx context.Context, input HookInput,
+				) (HookResult, error) {
+					return HookResult{Continue: true}, nil
+				}},
+			},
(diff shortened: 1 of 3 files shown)


Roasbeef:

Pointer is correct and the doc comment explains why in the way that matters: nil is not false. A CLI predating the field ignored `hooks` on every repeated initialize *without reporting it*, so a nil after a `Reinitialize` that carried hooks is the absence of evidence, not evidence of absence. Collapsing the two would let a caller conclude its hooks took effect when they may have been silently dropped.

Pointing the comment at `Reinitialize` is the right placement — that is the only call where a `false` is actionable, and the failure it describes (a client reattaching to a session another client configured, whose callbacks then never fire) has no other signal.

The integration test is live rather than a stub, with the assertion conditional on presence: this SDK owns the CLI's stdin, so a reported value must be true. It skips cleanly on 2.1.222 today and begins asserting when the runner catches up, following the `TestIntegrationAssistantContextUsage` precedent. Ran it locally against the live CLI to confirm the skip path is reached rather than an error path.

CI green.

VERDICT: APPROVE

PR #214, opened 2026-08-24
options: add v0.3.241 Settings parity fields

Roasbeef:

Six additions to `Settings`.

### `syncClaudeAiSkills` (L5392)

Turns off syncing of the skills enabled on claude.ai. **Only `false` is
honored** — the feature is enabled server-side per account, so setting `true`
does not turn it on early. That makes `false` the one value `omitempty` would
otherwise swallow, which is why it lands as a `*bool`; there is a dedicated
test pinning that.

The effect depends on which file carries it, and the asymmetry is worth
knowing before setting it:

- **user / managed settings** — destructive. Previously synced skills under
  `~/.claude/skills/synced` stop running, are hidden from later sessions, and
  move to `~/.claude/skills/.trash` at next launch. Deleted after
  `cleanupPeriodDays`, and re-*downloaded* rather than restored if you revert.
- **`.claude/settings.local.json` / `--settings`** — scoped and reversible.
  Downloads stop and synced skills are blocked for that workspace or
  invocation only. Nothing moves.

Not read from project settings at all.

### `worktree.location` (L5766)

Where Claude Code Desktop puts the worktrees of SSH sessions running on this
machine, instead of `<project>/.claude/worktrees`.

Called out in the comment because it is a trap for this SDK's users: **the CLI
does not read it.** Not `--worktree`, not `EnterWorktree`, not agent
isolation. Setting it from an SDK-driven session does nothing.

### `keybindingFlavor` (L7640)

`classic` (default) or `readline`, as a named type with two consts. The
difference is mostly Ctrl+W and what counts as a word — `readline` treats a
word as a run of letters and digits so punctuation separates words, where
`classic` uses Unicode word segmentation and keeps `foo_bar` and `3.14`
whole.

### `autoContinueAtUsageLimit` (L7670)

Wait out a claude.ai usage limit and continue automatically. When off, the
limit dialog offers the wait as a choice.

### `modelSettings` (L7415)

Per-model settings keyed by canonical model name — the per-model twin of the
existing top-level `EffortLevel`. Note the ceiling differs and the comment
says so: a persisted per-model effort tops out at `xhigh`, where the init
message's applied effort (see #207) also admits `max`.

### `spellcheck` (L7381)

New `SettingsSpellcheck` block: `enabled`, `checker`, `language`, `color`.
Read from user, flag and managed settings only — the whole block from the
highest-precedence of those applies, and it is ignored in
`.claude/settings.json` and `.claude/settings.local.json`.

Both `modelSettings` entries and the `spellcheck` block carry open index
signatures in TS. They land as plain structs per repo convention, with the
comment noting the CLI may add keys these drop on a round trip.

### Testing

Round-trip and omitempty tests over all six, plus the `syncClaudeAiSkills:
false` case on its own.

Integration test is **live**, as a new `v0_3_241_fields` subtest on
`TestIntegrationSettingsOptions` — it feeds all six through `--settings` to a
real session and asserts the round trip.

Probed the bare binary first, per the known failure mode where an unknown
*nested source-union* member fails managed-tier schema validation and stalls
the handshake for the full timeout instead of erroring. These are all
top-level scalars, maps or plain nested objects, so none of them trip it —
CLI 2.1.222 runs a `--settings` payload carrying all six to completion.

```
--- PASS: TestIntegrationSettingsOptions/v0_3_241_fields (1.35s)
```

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2092,6 +2092,52 @@ exec "$real_cli" "$@"
 		assert.Equal(t, want.Env, got.Env)
 	})
 
+	// The v0.3.241 fields are all top-level scalars, maps or plain nested
+	// objects — no source-union members — so the installed CLI accepts them
+	// rather than failing schema validation and stalling the handshake the
+	// way an unknown union variant does. Verified against the bare binary
+	// before landing this: CLI 2.1.222 runs a --settings payload carrying all
+	// six to completion.
+	t.Run("v0_3_241_fields", func(t *testing.T) {
+		syncOff := false
+		autoContinue := true
+		spellcheckOn := true
+
+		want := Settings{
+			SyncClaudeAiSkills:       &syncOff,
+			KeybindingFlavor:         KeybindingFlavorReadline,
+			AutoContinueAtUsageLimit: &autoContinue,
+			Worktree: &SettingsWorktree{
+				Location: "~/src/worktrees",
+			},
+			ModelSettings: map[string]SettingsModel{
+				"claude-opus-4-7": {EffortLevel: EffortXHigh},
+			},
+			Spellcheck: &SettingsSpellcheck{
+				Enabled:  &spellcheckOn,
+				Checker:  "auto",
+				Language: "en_GB",
+				Color:    "ansi256(203)",
+			},
+		}
+
+		argv := runWithArgvCapture(t, WithSettings(want))
+		var got Settings
+		require.NoError(t, json.Unmarshal(
+			[]byte(argValue(t, argv, "--settings")), &got))
+
+		require.NotNil(t, got.SyncClaudeAiSkills)
+		assert.False(t, *got.SyncClaudeAiSkills,
+			"only false is honored upstream, so it must survive the round trip")
+		assert.Equal(t, KeybindingFlavorReadline, got.KeybindingFlavor)
+		require.NotNil(t, got.AutoContinueAtUsageLimit)
+		assert.True(t, *got.AutoContinueAtUsageLimit)
+		require.NotNil(t, got.Worktree)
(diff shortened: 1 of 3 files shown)


Roasbeef:

Six fields, all additive, all round-tripped and omitempty-checked.

`SyncClaudeAiSkills` as a `*bool` is the one that needed thought: only `false` is honored upstream, which makes it precisely the value a plain `bool` with omitempty would drop on the floor. There is a dedicated test pinning that rather than leaving it to the shared round-trip case.

Two doc details I checked against the TS source and consider worth having carried over:

- The `SyncClaudeAiSkills` scope asymmetry is real and consequential — in user or managed settings a `false` moves previously synced skills to `.trash` for eventual deletion and re-download, while in `--settings` or `settings.local.json` it only blocks and hides them. Same key, materially different blast radius.
- `Worktree.Location` is desktop-only. The CLI reads it for neither `--worktree`, `EnterWorktree`, nor agent isolation, so for users of *this* SDK specifically it is inert. Better stated than discovered.

`ModelSettings` and `Spellcheck` land as typed structs despite TS declaring open index signatures on both, consistent with how this repo has handled that before, and the comments note keys may be dropped on a round trip.

Integration coverage is live, not a stub. Probed the bare binary first given the known failure mode where an unknown nested source-union member fails managed-tier schema validation and stalls the handshake for the full timeout instead of erroring — these are all top-level scalars, maps or plain nested objects, and CLI 2.1.222 runs a payload carrying all six to completion. The `v0_3_241_fields` subtest passes against the live CLI.

Rebased onto #210, which appended to the same EOF region of `options_test.go`. Resolved by keeping both blocks; all four affected tests re-verified locally.

CI green across all five build matrix entries.

VERDICT: APPROVE

PR #215, opened 2026-08-24
options: document headersHelper on marketplace sources

Roasbeef:

`sdk.d.ts` v0.3.241 adds `headersHelper` to the url marketplace source, and
`headers` + `headersHelper` to inline catalog entries — repeated across all
four marketplace settings keys (`extraKnownMarketplaces`,
`additionalMarketplaces`, `strictKnownMarketplaces`, `allowedMarketplaces`),
which is most of why the raw diff looked bigger than it is.

`SettingsMarketplaceSource` is an open `map[string]interface{}`, so both keys
are already representable. **This PR is the contract, not a new field** — the
same treatment `skipLfs` and `sha256` got in earlier cycles. It adds the
missing `SettingsMarketplaceSourceURL` const (the url variant had none) and
documents what these keys actually do.

### What is worth documenting

The two helpers look identical and are not:

- **url source** — re-run on **every later refresh** of the marketplace. It
  has to stay callable for the life of the config, not just at install time.
  It also runs from a *fixed* directory (the Claude config home, never the
  session's), so a relative path silently resolves somewhere the caller did
  not intend. Bare command on PATH, or absolute.
- **inline catalog entry** — runs **only** when a user explicitly installs or
  updates that plugin, and is scoped to that entry's `archive` download.

Two more asymmetries on catalog entries, both easy to get wrong:

- An entry declared in a settings file does **not** need `strict: false` the
  way a manifest catalog entry does — a settings file has no manifest fields
  to inline.
- A declaration in **project** settings is not operator-authored, so
  request-routing and client-identity header names stay filtered there, even
  though the identical declaration in user or managed settings passes.

### Also

Trims the `additionalMarketplaces` alias comment. Upstream reworded L6152 to
drop the claim that a client predating the alias would have its settings sync
re-upload the file as declaring no marketplaces; the Go comment still carried
it.

### Testing

Round-trip tests for a url source carrying `headers` + `headersHelper`, and
for an inline catalog entry carrying both, plus the const value. Since the
descriptor is an open map, what these assert is that the documented keys
survive marshal/unmarshal intact — which is the actual contract a caller
depends on here.

No integration test: this changes no wire behavior the SDK controls. Feeding a
real marketplace to the CLI is also the case that stalls the handshake on
schema mismatch, which `TestIntegrationSettingsMiscFields` already defers.

### DIFF
--- a/options.go
+++ b/options.go
@@ -372,9 +372,10 @@ type Settings struct {
 	// extraKnownMarketplaces when it updates the file.
 	//
 	// Prefer ExtraKnownMarketplaces while older clients still share the same
-	// settings: a client predating the alias ignores it outright, and its
-	// settings sync then uploads a file that declares no marketplaces at all.
-	// Mirrors sdk.d.ts v0.3.233 L6023.
+	// settings: a client predating the alias ignores it outright. Mirrors
+	// sdk.d.ts v0.3.241 L6152, which dropped the earlier claim that such a
+	// client's settings sync re-uploads the file as declaring no
+	// marketplaces.
 	AdditionalMarketplaces map[string]SettingsMarketplace `json:"additionalMarketplaces,omitempty"`
 	// AllowedMarketplaces is read exactly as if it were spelled
 	// StrictKnownMarketplaces, and like it is honored only from managed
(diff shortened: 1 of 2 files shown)


Roasbeef:

Correctly scoped as documentation plus one missing const, not a new field. `SettingsMarketplaceSource` is an open map, so both keys already ride the wire — the same treatment `skipLfs` and `sha256` got in earlier cycles. Inventing a typed shape here would fight the union rather than model it.

What justifies the PR is that the two helpers look interchangeable and are not. The url source's helper is re-run on **every** marketplace refresh and executes from the Claude config home rather than the session's directory, so a relative path silently resolves somewhere the caller did not intend. The catalog entry's runs only on explicit install or update. Getting those backwards yields either a stale token or a helper invoked far more often than expected.

The two catalog asymmetries are also worth carrying: a settings-declared entry does not need `strict: false` (a settings file has no manifest fields to inline), and a project-settings declaration still has request-routing and client-identity header names filtered, because it is not operator-authored.

Tests assert the documented keys survive marshal/unmarshal on both shapes, which is the real contract for an open map. The `additionalMarketplaces` comment trim tracks an upstream reword, not a behavior change.

Rebased across #210 and #214, which both appended to the same EOF region of `options_test.go`. Rebuilt the file from main plus this branch's append and re-ran all thirteen affected tests locally.

CI green.

VERDICT: APPROVE

PR #216, opened 2026-08-24
docs: refresh TypeScript SDK parity to v0.3.241

Roasbeef:
Part of #206 (TypeScript SDK v0.3.241 parity). Closes the cycle.

Brings the README parity section up to v0.3.241: the header version, a section
for what this cycle added, the PR list, and what was deferred.

The prior cycle's "Deferred this cycle" heading is retitled "Deferred in
v0.3.233" — with a second cycle's deferrals below it, "this cycle" had become
ambiguous about which one it meant.

Nothing beyond documentation. The nine feature PRs (#207–#215) carry the code.

### Deferred in v0.3.241

- **`sdk-tools.d.ts` churn** (185 diff lines, the Artifact/read/watch tool input
  unions) — outside the curated `tool_inputs.go` subset. Standing deferral.
- **`bridge.d.ts`** — the `onStopTask` callback, and the terminal-reason split
  where `request_rejected` gains `status: number` and `source?: 'origin' |
  'nonorigin_cf' | 'nonorigin_other'` while `malformed_response` becomes its own
  variant. Deferred in v0.3.220 and v0.3.226 for the same reason: no Go bridge
  transport exists.
- **Two upstream removals that are no-ops here** — `ExitReason` /
  `EXIT_REASONS` lost `bypass_permissions_disabled`, and
  `SDKControlGetPlanRequest` / `SDKControlGetWorkspaceDiffRequest` were deleted
  outright (gone from `sdk.mjs` too, not just the types). The Go SDK never
  modeled either surface.

### Cycle notes

The `sdk.d.ts` diff was 815 lines but only **69** non-comment — most of the
bulk was doc expansion on types already modeled here. `sdk.mjs` emits an
identical `--flag` set, so there was no transport work at all this cycle, and
nothing breaking landed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ stdin/stdout, giving you access to Claude's tool use, extended thinking,
 session management, and hook system.
 
 This repository tracks the official TypeScript Agent SDK surface through the
-v0.3.233 catchup work, using Go idioms where the API shape differs.
+v0.3.241 catchup work, using Go idioms where the API shape differs.
 
 ```mermaid
 flowchart TB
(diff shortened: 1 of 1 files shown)


Roasbeef:
Reviewed as orchestrator for the v0.3.241 catchup. Closes the cycle.

Documentation only; the nine feature PRs carry the code. Verified the PR list against the merged state — #207 through #215 are all MERGED with titles matching what the README claims each contributed.

Retitling the previous cycle's `Deferred this cycle` to `Deferred in v0.3.233` is the right call now that a second cycle's deferrals sit below it. `this cycle` had become genuinely ambiguous about which one it referred to.

The deferral list is accurate and each entry gives a reason rather than just a name: `sdk-tools.d.ts` is the standing curated-subset exclusion, `bridge.d.ts` has no Go transport to attach to (consistent with v0.3.220 and v0.3.226), and the two upstream removals are recorded as explicit no-ops rather than silently omitted — worth having on record, since `ExitReason` losing a member and two control requests being deleted outright would otherwise look like an oversight to a future reader diffing the same range.

CI green.

VERDICT: APPROVE

Issue #217, opened 2026-08-31
Parity with TypeScript Agent SDK v0.3.251


Roasbeef:
Tracking issue for the weekly TypeScript SDK catchup. Prior shipped parity: **v0.3.241** (#206, PRs #207–#216).

`sdk.d.ts` diffed 662 lines against v0.3.241; roughly 60 of those are non-comment. `sdk.mjs` emits an identical `--flag` set, so there is no transport or argv work this cycle, and `agentSdkTypes.d.ts` is byte-identical.

The headline is two new hook events — `PreModelSwitch` and `PostModelSwitch` — the first new hook pair since `DirectoryAdded` in v0.3.220. Beyond that the cycle has a theme: prompt-cache cost visibility. The model-switch hook inputs, the four new `SessionStart` fields, `Settings.promptCacheTtl` / `subagentPromptCacheTtl`, and `ModelUsage.costBasis` all exist so a host can price a model switch or a session resume in cache-write dollars before committing to it.

Everything is additive except `Settings.spinnerTipsOverride`, where `tips` went required-to-optional and its element widened to a union. That lands non-breaking — see PR-10.

## PRs

- [ ] 01 — `PreModelSwitch` + `PostModelSwitch` hook events
- [ ] 02 — per-server timeout for SDK MCP servers (`sdkMcpServerConfigs`)
- [ ] 03 — `ModelUsage.costBasis`
- [ ] 04 — `Options.perTaskStopAffordance`
- [ ] 05 — `user_message_uuid` on assistant / partial / result-error
- [ ] 06 — `queued_turn_count` on both result shapes
- [ ] 07 — `ambient` on task_started / task_notification / background_tasks_changed
- [ ] 08 — `default_to_no` on the `can_use_tool` control request
- [ ] 09 — `SessionStart` hook input cache context (+ two pre-existing gaps)
- [ ] 10 — Settings scalar parity + `spinnerTipsOverride` reshape
- [ ] 11 — `Settings.modelPicker` + `Settings.modelPricing`
- [ ] 12 — parity docs refresh

## Deferred

- **`bridge.d.ts` `onSetModel`** — gained `notices?: string[]` and a `Promise<...>` return variant. The Go SDK does not model the claude.ai bridge surface at all; deferred every cycle since v0.3.233.
[...]

PR #218, opened 2026-08-31
hooks: add PreModelSwitch and PostModelSwitch events

Roasbeef:

A model switch throws away the outgoing model's warm prompt cache, and re-warming it against the new model is not free. v0.3.251 adds a hook pair around that moment so a host can see the bill before it is incurred. Both inputs carry the same context: `from_model`, `to_model`, `requested_model`, `context_tokens`, `prompt_cache_warm`, `cache_ttl`, `estimated_cache_write_usd`, and `pricing`.

The two differ in what they can do about it. `PreModelSwitch` is vetoable on the same contract as `PreToolUse` — `deny` cancels the switch, `ask` defers to the user (a headless session refuses instead), `allow` proceeds and skips the interactive cache-miss confirm. `PostModelSwitch` is a report and takes only `additionalContext`.

They also differ in which switches they see. `Pre` admits `command | picker | sdk`; `Post` adds `auto | resume`. A switch the CLI performs on its own — a refusal fallback, restoring a model on resume — has no pre-phase to veto, so those two values only ever surface after the fact. That asymmetry is the reason the shared fields sit in an embedded `modelSwitchContext` while `Source` is spelled separately on each type.

## permissionDecision gets a typed path

There was no typed way to return `permissionDecision` before this — `PreToolUse` callers reached it through the raw `HookSpecificOutput` map. Adding a field that worked only for `PreModelSwitch` would have been the odd shape, so it lands as `HookResult.PermissionDecision` + `PermissionDecisionReason`, honored on both pre-phase hooks and silently dropped elsewhere. It composes with an explicit `HookSpecificOutput` map the way `AdditionalContext` already does: the typed value overwrites its own key and leaves the caller's others alone.

The value sets are not identical and the doc comment says so. `HookPermissionDecision` carries all four upstream constants including `defer`, but `defer` is `PreToolUse`-only — there is no deferred model switch to hand off to.

## Also here

- `CacheTTL` (`5m` / `1h`) is defined here and reused by `Settings.promptCacheTtl` / `subagentPromptCacheTtl` in a later PR of this series rather than being spelled twice.
- `ModelPricingBasis` (`configured` / `catalog` / `default`) — `configured` is what `Settings.modelPricing` produces, and it is the same distinction `ModelUsage.costBasis` reports. Both land later in the series.
- `RequestedModel` is `*string`. Upstream types it `string | null`, but nothing consumes the difference between an absent field and an explicit null here, so a plain pointer is honest and the nullable-value dance is unnecessary.

## Testing

Both `protocol.go` hook-input builders are updated — they are separate field-mapping switches, and a field added to only one is silently empty in the other. Unit tests cover each builder independently plus the three `permissionDecision` response paths (emitted, dropped on an unsupported hook, composed with an explicit map).

`TestIntegrationModelSwitchHooks` drives a real switch through `Stream.SetModel` and asserts `source == "sdk"`. It skips on CLI 2.1.222, which does not emit either event yet.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


Roasbeef:
## Review — PR-01, v0.3.251 catchup

Reviewed against `scratch/ts-sdk-0.3.251/package/sdk.d.ts` L853, L872, L874, L2325, L2444, L8261.

**Field-set parity.** Both input types carry all eight upstream fields with matching wire names, and the `source` value sets are correctly asymmetric — Pre stops at `sdk`, Post adds `auto` and `resume`. Verified against L2331 and L2450 directly rather than from the diff, per the lesson from PR-04 of the v0.3.168 cycle.

**Dual builder.** Both `handleHookCallback` (L552 area) and `handleSDKHookCallback` (L1069 area) are updated, and each has its own test rather than one test standing in for both. This is the failure mode that bit #173.

**Embedded unexported struct.** `modelSwitchContext` is embedded by value, not by pointer. That matters: `encoding/json` promotes the exported fields of an unexported embedded struct on both marshal and unmarshal, but cannot allocate a nil unexported embedded *pointer* on decode. Round-tripped a populated `PreModelSwitchInput` to confirm all eight promoted fields survive.

**permissionDecision precedence.** The new block runs after the `Modify` auto-translate, so an explicit `PermissionDecision` overwrites the hardcoded `"allow"` that PreToolUse's `Modify` path emits. That is the right precedence — a typed field the caller set beats one the SDK inferred. `PermissionDecisionReason` is omitted when empty rather than emitting `""`.

**Scope.** Extending `permissionDecision` to `PreToolUse` is beyond the literal v0.3.251 delta. It is the right call anyway: a typed field honored on exactly one of the two hooks that upstream defines it for would be the strange shape, and the alternative was leaving `PreToolUse` callers on the raw map indefinitely. The doc comment carries the `defer` asymmetry, which is the part that would otherwise bite someone.

**Coverage.** `TestIntegrationModelSwitchHooks` skips on CLI 2.1.222 — the CLI does not emit either event for `set_model` yet. Confirmed by running it locally, not assumed. The test is a real one that will start exercising the path once the CLI catches up, not a permanent stub.

Nothing to change. All matrix entries green, Lint clean, `golangci-lint run` clean locally.

VERDICT: APPROVE

PR #219, opened 2026-08-31
mcp: add per-server tool-call timeout for SDK MCP servers

Roasbeef:

Remote MCP servers have carried a per-server `timeout` since v0.3.150 — `MCPServerConfig.Timeout` already models it. The in-process ones had no equivalent and fell through to `MCP_TOOL_TIMEOUT` or the CLI default. v0.3.251 closes the gap with three coupled additions: `CreateSdkMcpServerOptions.timeout` (L539), `McpSdkServerConfig.timeout` (L1095), and `SDKControlInitializeRequest.sdkMcpServerConfigs` (L3847).

## How it reaches the CLI

Not through the name list. `sdkMcpServers` stays a bare `[]string`, and the new `sdkMcpServerConfigs` map sits beside it keyed by the same names. Reading `sdk.mjs` confirms the CLI builds that map from a `flatMap` over the registered servers, keeping only those with a defined timeout, and drops the key entirely when the result is empty. This mirrors that: the map is allocated lazily on the first server that sets a timeout and stays `nil` otherwise. A missing key and `{}` mean the same thing to the CLI, so sending `{}` is just noise.

## Units, and two constraints worth stating

Milliseconds — stated outright at L1093, not inferred from the bundle. The doc comment also carries the two behaviors that would otherwise read as bugs:

- Values below 1000ms are silently ignored and fall through to `MCP_TOOL_TIMEOUT`. So is anything non-positive or non-integer; `sdk.mjs` validates with exactly that predicate.
- The value is read when the server is registered. Changing it on a server the CLI already knows about has no effect until the server is removed and re-added — `sdk.mjs` logs a warning on precisely that case.

`McpServer.Timeout()` is exported alongside the existing `Instructions()` accessor, since the protocol layer needs to read it back off a registered server.

## Testing

Two cases in the existing `TestProtocolInitializeOptions` table: one server with a timeout produces both `sdkMcpServers` and `sdkMcpServerConfigs`, one without produces the name list and no config key at all.

`TestIntegrationSdkMcpServerTimeout` is a real behavioral test rather than an argv assertion. It registers a tool that sleeps for a minute behind a two-second per-server timeout and asserts the turn completes anyway. That can only happen if the field reached the CLI and was honored — it passes in ~19s against CLI 2.1.222.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2763,3 +2763,81 @@ func TestIntegrationModelSwitchHooks(t *testing.T) {
 			"a post-switch report must name two different models")
 	}
 }
+
+// TestIntegrationSdkMcpServerTimeout registers an in-process MCP server with a
+// short per-server timeout and a tool that sleeps well past it, then asserts
+// the CLI cuts the call off rather than waiting for the tool. This exercises
+// sdkMcpServerConfigs on the initialize request, added in TS SDK v0.3.251.
+func TestIntegrationSdkMcpServerTimeout(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	const (
+		toolTimeoutMs = 2000
+		toolSleep     = 60 * time.Second
+	)
+
+	type NoArgs struct{}
+
+	timeout := toolTimeoutMs
+	server := CreateMcpServer(McpServerOptions{
+		Name:    "slowpoke",
+		Timeout: &timeout,
+		Tools: []ToolRegistrar{
+			Tool("stall", "Stall for a long time",
+				func(ctx context.Context, args NoArgs) (ToolResult, error) {
+					select {
+					case <-time.After(toolSleep):
+						return TextResult("finished"), nil
+					case <-ctx.Done():
+						return TextResult("cancelled"), nil
+					}
+				},
+			),
+		},
+	})
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("You have one tool, mcp__slowpoke__stall. When asked "+
+			"to run it you MUST call it. Do not explain, do not refuse, do "+
+			"not ask questions. Call it exactly once, then stop."),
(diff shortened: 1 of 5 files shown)


Roasbeef:
## Review — PR-02, v0.3.251 catchup

Reviewed against `sdk.d.ts` L539, L1095, L3847 and the `sdk.mjs` initialize builder.

**Wire shape.** The `sdk.mjs` construction is `flatMap` over the registered servers keeping only those with a defined timeout, then `Object.fromEntries` guarded by a length check that leaves the key `undefined` when empty. The Go side reproduces that exactly with lazy map allocation, and the second table case pins the omission rather than leaving it to inspection. Emitting `{}` would not have broken anything, but pinning the absence keeps the two implementations comparable byte for byte.

**Units.** Milliseconds, confirmed from the L1093 doc text rather than reverse-engineered from the bundle — the `??600000` constants nearby belong to the vendored Anthropic API client, not to MCP, and would have been the wrong thing to read.

**The two silent-drop behaviors are documented.** Sub-1000ms values and the registration-time read are both the kind of thing that turns into a bug report six months out. `sdk.mjs` validates with `typeof e === "number" && Number.isInteger(e) && e > 0`, so non-positive and non-integer values are dropped on the same path; the comment covers those too.

**Accessor.** `Timeout()` matches the existing `Instructions()` shape. The nil check on `server` in the initialize loop is defensive against a nil map value rather than a real path, but the loop already dereferences, so it costs nothing.

**Integration test.** This is the good kind: a 60s tool behind a 2s timeout, asserting the turn finishes at all. It cannot pass unless the field crossed the wire and the CLI honored it, which is exactly the divergence class the integration suite exists for. Verified locally at ~19s against CLI 2.1.222. The first draft skipped because the model declined the tool call; the prompt was tightened until the call was reliable rather than leaving a test that silently skips.

VERDICT: APPROVE

PR #220, opened 2026-08-31
messages: record pricing provenance on ModelUsage

Roasbeef:

`ModelUsage.costUSD` has never said where its number came from. v0.3.251 adds `costBasis` (sdk.d.ts L1326) with three values:

- `list` — Claude Code's built-in list prices
- `managed` — the organization's managed-settings `modelPricing` rates or multiplier
- `unknown` — neither matched, so `costUSD` is the *default model's* rate standing in for a model nothing knew how to price

That last one is the interesting case. It is not a rounding concern; it means the figure is about a different model than the one it is filed under. And the `list` / `managed` split matters to anything that aggregates: summing across both produces a number that is not any organization's actual spend, and today a consumer has no way to notice it is doing that.

## canonicalModel and provider go in at the same time

Both have been on the wire since v0.3.220. That cycle added them to `ModelInfo`, which is a different type — `ModelUsage` was missed and has been silently dropping them since. The integration test confirms CLI 2.1.222 is already sending both:

```
claude-sonnet-4-5-20250929: canonical="claude-sonnet-4-5" provider="firstParty" basis="" cost=0.0278643
```

Same struct, same decode, and a diff-driven catchup will never surface them again on its own. Folding them in here rather than opening a separate PR for two fields nobody would otherwise find.

## Absence is not zero

All three are empty until this process has priced a request for the model — a `--resume` can defer that indefinitely — and on CLIs that predate them. Upstream says to read an empty `costBasis` as `list`, and the doc comment says so too rather than leaving callers to guess. `costBasis` is also overwritten per request rather than accumulated, so differencing cumulative `costUSD` across turns gives you that turn's basis; that is noted as well, since it is the non-obvious half.

## Testing

Round-trip with all three set, plus a decode of a payload carrying none of them that asserts the keys stay off the wire on re-marshal.

`TestIntegrationModelUsageCostBasis` runs live and passes. It hard-asserts `canonicalModel` and `provider` — they are on the wire today, so a silent regression there should fail rather than skip — and skips only the `costBasis` half, which CLI 2.1.222 does not emit yet.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2841,3 +2841,70 @@ func TestIntegrationSdkMcpServerTimeout(t *testing.T) {
 		"a per-server timeout of %dms must cut the call off well before the "+
 			"tool's own %s sleep", toolTimeoutMs, toolSleep)
 }
+
+// TestIntegrationModelUsageCostBasis asserts the pricing-provenance fields on
+// ModelUsage — costBasis is new in TS SDK v0.3.251, canonicalModel and provider
+// predate it and were never modeled here.
+func TestIntegrationModelUsageCostBasis(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("You are a helpful assistant. Be very brief."),
+		WithMaxTurns(1),
+	)
+	client, err := NewClient(opts...)
+	require.NoError(t, err)
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
+	defer cancel()
+
+	var result *ResultMessage
+	for msg := range client.Query(ctx, "Say OK.") {
+		if m, ok := msg.(ResultMessage); ok {
+			result = &m
+			break
+		}
+	}
+	require.NotNil(t, result, "expected a result message")
+	require.NotEmpty(t, result.ModelUsage, "expected per-model usage")
+
+	var sawBasis, sawCanonical bool
+	for model, usage := range result.ModelUsage {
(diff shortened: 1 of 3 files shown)


Roasbeef:
## Review — PR-03, v0.3.251 catchup

Reviewed against `sdk.d.ts` L1306–L1327.

**Value set and semantics.** Three constants, matching upstream. The `unknown` doc comment is the one that earns its keep: upstream says `costUSD` in that case is "a guess at the default model's rate", which means the number is about a *different model* than the key it sits under. Stating that is more useful than restating the enum.

**Per-request overwrite.** Upstream notes `costBasis` is overwritten per request like `canonicalModel`, so a consumer differencing cumulative `costUSD` per turn gets that turn's basis. That is the non-obvious half of the contract and it made it into the comment. Without it, a reader would reasonably assume the basis describes the whole accumulated figure.

**The gap-fill is justified by evidence, not assumption.** `canonicalModel` and `provider` went onto `ModelInfo` in v0.3.220 and `ModelUsage` was missed. The integration test log proves the CLI has been sending both here all along:

```
canonical="claude-sonnet-4-5" provider="firstParty"
```

A diff-driven catchup only ever sees what changed, so a field the Go side missed in a prior cycle is invisible forever unless someone reads the whole type. Correct to fold in rather than defer.

**Absent-vs-zero handling.** All three take `omitempty` against a struct whose other fields deliberately do not, which is right — they are optional upstream and the rest are not. `TestModelUsagePricingProvenanceAbsent` pins both directions: an old-CLI payload decodes empty, and re-marshalling does not invent the keys.

**Integration test shape.** Hard-asserts the two fields that are live today and skips only the one the CLI does not emit yet. That is the right split — a permanent skip over the whole test would have hidden a regression in the two that work.

VERDICT: APPROVE

PR #221, opened 2026-08-31
options: add perTaskStopAffordance to the initialize request

Roasbeef:

`Options.perTaskStopAffordance` (sdk.d.ts L1667) and the matching `SDKControlInitializeRequest.perTaskStopAffordance` (L3885).

## It is not a display hint

The name reads like a UI capability flag, but the field decides what an interrupt does:

- **Declared** — an interrupt on an open-input (interactive stream-json) session aborts only the current turn. Background agents and workflows keep running, and the user stops them one at a time through the consumer's own control, wired to `stop_task`.
- **Absent** — the interrupt kills background tasks.

The kill is the default on purpose. A consumer that cannot render a per-task stop control has no way to stop a runaway task short of ending the session, so the CLI fails closed rather than sparing something the user cannot reach. That reasoning is in the doc comment, because "absent means kill" looks arbitrary without it.

One exception: a one-shot run — the string-prompt form, or `-p` — closes stdin, and with stdin closed a `stop_task` control could never be delivered. Hold-back tasks are killed at the held-result release regardless of the declaration.

It is `*bool`, not `bool`. The field selects between two real behaviors, so an explicit `false` is worth being able to send rather than collapsing into "unset".

## Shape

Follows `SupportedDialogKinds`, which it sits directly beside in both the options type and the initialize request, and shares its first-attached-client-wins rule on multi-client sessions.

## Testing

Three cases in `TestProtocolInitializeOptions`: true wired through, explicit false wired through as `false` rather than dropped, unset omits the key.

`TestIntegrationPerTaskStopAffordance` declares the field and completes a turn. That is a weaker assertion than the last two PRs and the test says so — an initialize carrying an unrecognized key is not automatically benign (an unknown nested member can stall the handshake outright), so completing a turn is the cheap proof this one is understood. Observing the behavior it selects needs a long-running background task plus an interrupt; that is filed in `INTEGRATION-FOLLOWUPS.md` alongside a note for PR-01's model-switch hook test.


Roasbeef:
## Review — PR-04, v0.3.251 catchup

Reviewed against `sdk.d.ts` L1667 and L3885.

**The comment does the real work here.** The wire change is one `*bool`; the value is in stating that this is not a display capability flag. It selects between "interrupt aborts the turn and spares background tasks" and "interrupt kills them", and the default is the kill. Without the fail-closed rationale — a consumer that cannot render a stop control has no way to reach a runaway task short of ending the session — "absent means kill" reads as an arbitrary choice, and someone eventually "fixes" it.

The closed-input exception is carried too. `-p` and the string-prompt form close stdin, `stop_task` can never be delivered, so hold-back tasks die at the held-result release no matter what was declared. That is exactly the kind of thing that produces a bug report against the SDK rather than against the CLI.

**`*bool` over `bool`.** Right call for a field that selects between two behaviors: an explicit `false` is a meaningful thing to send, and collapsing it into "unset" would be lossy. The middle table case pins that `false` reaches the wire as `false` rather than being dropped by `omitempty`, which is the failure mode a plain `bool` would have had.

**Placement.** Beside `SupportedDialogKinds` in both the options type and the initialize request, matching upstream's own ordering, and sharing its first-attached-client-wins semantics.

**Integration test honesty.** This is weaker than the previous two PRs' and the test says so in its own comment rather than implying more coverage than it has. Completing a turn only proves the CLI parses the key — but that is not nothing: an unrecognized nested member can stall the handshake outright rather than being ignored, so "the session still works" is a real signal here. The behavioral half is filed in `INTEGRATION-FOLLOWUPS.md`, along with a matching entry for PR-01's hook test.

**CI note.** The first `Test` run failed on `TestProtocolInitializeHookTimeout/timeout_zero_omitted` — a 1s-context deadline under `-race`, in a test this diff does not touch. Reproduced clean locally at `-race -count=5`; green on re-run. Same flake family as the `TestProtocolInitializeOptions` deadline seen in the v0.3.201 cycle.

Nothing to change.

VERDICT: APPROVE

PR #222, opened 2026-08-31
messages: echo user_message_uuid on every first reply frame

Roasbeef:

The success result has echoed the client uuid of the triggering send since v0.3.220. That is enough to reconcile a turn after it finishes, but not to bind a reply while it is still streaming. v0.3.251 closes that: `SDKAssistantMessage` (L3211) and `SDKPartialAssistantMessage` (L4655) now carry `user_message_uuid`, as does `SDKResultError` (L4799).

## The error result needed no new field

Go models success and error as one `ResultMessage`, so the existing field already covers the error case. What it did need was the comment fixed — it read `(success only; sdk.d.ts v0.3.220 L4300)`, which is no longer true and would have sent the next reader looking for a field that is already there. Called out here rather than quietly narrowing scope.

## It is a binding key, not a message attribute

The stamp marks a turn's *first* reply frame and nothing after it. In complete-message mode that is the first assistant message; with `IncludePartialMessages` the stamp normally rides the first non-ping stream event instead, though a turn that produces no stream events still stamps its first assistant message.

Everything else comes back empty: later frames of the same turn, subagent frames (`ParentToolUseID` set), synthetic and scheduled turns, sends that carried no client uuid, and older producers. Each shape's comment spells out its own exclusion list, because a consumer that reads this as "this frame belongs to that send" will find most frames unattributed and reasonably conclude the field is broken.

`PartialAssistantMessage` also picks up `ttft_ms`, which has been upstream since before this cycle and was never modeled here — same struct, same decode.

## Testing

Unit tests decode all three shapes through `ParseMessage` and pin that an absent field does not become an empty key on re-marshal.

`TestIntegrationUserMessageUUIDCorrelation` sends a real message with a client uuid. `Stream.Send` mints no uuid, so it goes through `Protocol.SendMessage` directly — there is nothing to echo otherwise. The result echo is **live on CLI 2.1.222 and hard-asserted**:

```
UserMessageUUID:550e8400-e29b-41d4-a716-4466554412ab
```

The first-reply-frame half skips — 2.1.222 leaves `UserMessageUUID` empty on assistant frames — and is filed in `INTEGRATION-FOLLOWUPS.md`. Splitting it this way keeps the part that works today from silently regressing behind a whole-test skip.


Roasbeef:
## Review — PR-05, v0.3.251 catchup

Reviewed against `sdk.d.ts` L3211, L4655, L4799.

**The error-result finding is handled correctly.** Go collapses `SDKResultSuccess` and `SDKResultError` into one `ResultMessage`, so upstream's third addition needs no field — only the stale `(success only)` annotation removed. Saying so in the PR body rather than dropping it from scope silently is the right instinct; a reader checking parity later would otherwise see three upstream adds and two Go fields and have to re-derive why.

**Framing.** Describing this as a binding key rather than a message attribute is the part that makes the comments useful. Upstream's own text buries it, but the operative fact is that the stamp lands on exactly one frame per turn — every later frame, every subagent frame, every synthetic turn comes back empty. A consumer reading it as provenance will conclude the field is broken on frame two. Each shape carries its own exclusion list rather than a shared hand-wave, which is right because they genuinely differ: assistant frames additionally exclude subagent frames, stream events do not.

**The complete-message vs partial-messages split** is also captured: with `IncludePartialMessages` the stamp normally moves to the first non-ping stream event, *except* on a turn that emits no stream events. That exception is what makes "assert one of the two" the only correct assertion shape, and the integration test does exactly that rather than pinning a frame type the CLI is free to choose.

**Gap-fill.** `ttft_ms` on `PartialAssistantMessage` predates this cycle and was never modeled. Same justification as PR-03's `canonicalModel` — a diff-driven catchup cannot see it.

**Integration test.** Goes through `Protocol.SendMessage` because `Stream.Send` mints no client uuid, which is the honest way to get something to correlate rather than testing a path that can never echo. The result echo is hard-asserted and verified live:

```
UserMessageUUID:550e8400-e29b-41d4-a716-4466554412ab
```

Only the first-frame half skips. Splitting the assertion instead of skipping the whole test keeps the working half from regressing unnoticed — same pattern as PR-03, and it is the right default for a field that lands in stages across CLI versions.

Worth noting for a future cycle: `Stream.Send` not accepting a client uuid means Go hosts cannot use this correlation at all through the public API. That is a separate surface (`submitMessage` options) and correctly out of scope here.

VERDICT: APPROVE

PR #223, opened 2026-08-31
messages: report the pending send backlog on results

Roasbeef:

`queued_turn_count` on both `SDKResultSuccess` (L4848) and `SDKResultError` (L4795). Go models the two as one `ResultMessage`, so it is one field.

## What it answers

A host draining a queue cannot currently tell "this run is finished" from "another result is already on its way" — a result message looks identical either way, and the only recourse is a timeout or bookkeeping the host maintains itself. `queued_turn_count` says it directly: greater than zero means at least one more turn and result follow without further input, barring cancellation.

## Nil, not zero

`*int` rather than `int`, because absent and zero are different claims:

- **Zero** — nothing is pending, *or* the session is ending. `end_session` and a shutdown latched mid-turn both discard the backlog, so zero can mean "drained" or "abandoned".
- **Nil** — this surface has no command queue at all, or the result is a fatal startup result, or the CLI predates the field.

Collapsing those would make a queueless surface indistinguishable from an idle one, which is precisely the distinction a host draining a queue needs.

## It bounds, it does not predict

The field counts *pending sends*, not remaining results, and queued sends may coalesce into fewer turns. System-generated queue entries are excluded. The name invites the stronger reading — "three more results are coming" — so the comment says what it actually guarantees.

## Testing

Three unit cases: a reported backlog, an explicit zero surviving the round trip as `"queued_turn_count":0` rather than being dropped by `omitempty`, and an absent field staying absent.

`TestIntegrationQueuedTurnCount` builds a real backlog — two sends issued before the first turn resolves — and asserts the first result reports it and the last reports zero. It skips on CLI 2.1.222, which omits the field even with the queue genuinely non-empty:

```
result 1: queued=absent
result 2: queued=absent
```

The backlog is constructed rather than assumed, so the test starts asserting the moment the CLI catches up. Filed in `INTEGRATION-FOLLOWUPS.md`.

Kept separate from #222 despite touching the same struct: turn-correlation identity and queue depth are different surfaces. Rebased on it.


Roasbeef:
## Review — PR-06, v0.3.251 catchup

Reviewed against `sdk.d.ts` L4795 and L4848.

**Pointer is load-bearing here.** Zero and absent carry genuinely different information: zero means the queue drained *or* the session is ending (`end_session` and a mid-turn shutdown both discard the backlog), while nil means there is no command queue on this surface at all. A plain `int` with `omitempty` would have merged the two and also silently dropped a real zero on re-marshal — the second unit case pins exactly that, asserting `"queued_turn_count":0` survives.

**The bounds-not-predicts distinction is worth the paragraph.** Upstream is explicit that queued sends may coalesce into fewer turns and that system-generated entries are excluded, so the count is an upper bound on what follows, not a countdown. The field name reads like a countdown. Left undocumented this becomes a host that waits for N more results and hangs.

**Integration test constructs the condition it needs.** Two sends issued before the first turn resolves is a genuine backlog, not an assumed one, so the skip is honest: the queue really was non-empty and 2.1.222 still reported nothing. That means the test flips to asserting the moment the CLI ships the field, with no further work. Contrast with a test that skips because it could not build the precondition — that one would need rewriting later.

**Split from #222.** Same struct, different surface, landed serially with a rebase rather than in parallel. Correct given both append to `integration_test.go` at EOF, which is the conflict shape that bit the v0.3.150 cycle.

Minor: `derefInt` exists only for a log line. It is three lines and confined to the integration build tag, so it stays.

VERDICT: APPROVE

PR #224, opened 2026-08-31
messages: add the ambient marker to task frames

Roasbeef:

`ambient` lands on three task shapes: `SDKTaskStartedMessage` (L5088), `SDKTaskNotificationMessage` (L5030), and the task summaries inside `SDKBackgroundTasksChangedMessage.tasks[]` (L3283). It marks housekeeping the CLI does not surface as user work, and hosts should keep those out of activity indicators.

## It also corrects a gloss that is now wrong

Both `SkipTranscript` fields in `messages.go` were annotated `// Ambient task marker`. That was a fair reading when no such field existed on the wire. It is not one now, and the two are not synonyms:

- `skip_transcript` — hide this from the inline transcript; it may still appear in a tasks panel.
- `ambient` — this is not user work at all. Covers **every** `skip_transcript` task, **plus** auto-started live-update watchers.

So `ambient` is the strictly broader set, and they drive different consumer behavior. A host that keeps treating them as the same thing will either surface watchers as activity or drop real tasks out of its panel. Left alone, the old comment actively points at the wrong conclusion now that the real field exists, which is why it is fixed here rather than in the docs PR.

`BackgroundTask` picks the field up too — that is the one message enumerating every live task, so it is where the ambient/real split most needs to be visible.

## Testing

Four unit cases. The `task_started` case deliberately asserts `ambient: true` with `skip_transcript` absent — that is the watcher shape, and it is the case that proves the two decode independently rather than one standing in for the other. `background_tasks_changed` mixes a marked and an unmarked task in one payload.

`TestIntegrationAmbientTaskMarker` spawns a real Explore subagent and inspects the frames. CLI 2.1.222 stamps neither marker:

```
task_started a4c8748899c1650ba: ambient=<nil> skip_transcript=<nil>
```

so it skips — but the task was genuinely spawned first, so the precondition is real and the assertion (a subagent the user asked for must not be marked ambient) starts running as soon as the CLI ships the field. Filed in `INTEGRATION-FOLLOWUPS.md`.

`ambient` appears zero times in `sdk.mjs`, consistent with it being inbound-only from the CLI.


Roasbeef:
## Review — PR-07, v0.3.251 catchup

Reviewed against `sdk.d.ts` L3283, L5030, L5088.

**The comment fix is the important part of this diff.** `// Ambient task marker` on `SkipTranscript` was correct-by-approximation while `ambient` did not exist. Now that it does, the old annotation points a reader at the wrong field for the wrong reason. Upstream's own text makes the containment explicit — ambient is every `skip_transcript` task *plus* auto-started watchers — and the two ask consumers for different things. Fixing it here rather than deferring to the docs PR is right: a wrong comment beside a newly added field is worse than a wrong comment alone.

**Independent decode is actually tested.** The `task_started` case asserts `ambient: true` with `skip_transcript` absent, which is the watcher shape and the only case that proves the two fields do not shadow each other. A test that set both together on every shape would have passed identically with a buggy implementation that aliased them.

**`BackgroundTask` inclusion.** Easy to miss — it is a nested anonymous type upstream and a named struct here, so the diff hunk attributes it to the enclosing message. Correct to include; it is the one frame that enumerates every live task, so it is where a host most needs the split.

**Integration test precondition is real.** The Explore subagent genuinely spawned before the skip fired:

```
task_started a4c8748899c1650ba: ambient=<nil> skip_transcript=<nil>
```

That distinguishes "the CLI does not send this yet" from "the test could not construct the situation". The assertion it will run — a user-requested subagent must not be marked ambient — is the one that would catch an upstream over-tagging regression, which is the failure mode that actually matters for a field driving activity indicators.

Consistent with `ambient` appearing zero times in `sdk.mjs`: inbound only, no dispatch path to cover.

VERDICT: APPROVE

PR #225, opened 2026-08-31
protocol: carry default_to_no on permission asks

Roasbeef:

`SDKControlPermissionRequest.default_to_no` (sdk.d.ts L4079). Some asks must not be approvable by a stray keystroke — the CLI's own prompt opens on the decline option and takes no digit shortcut. This tells SDK hosts the same thing, so they stop pre-selecting approve on an ask the CLI itself treats as decline-by-default.

## Both permission paths, and a gap found on one of them

`can_use_tool` reaches the SDK two ways: as a legacy `ControlRequest` with a `Payload` map, and as the SDK-format `SDKControlRequest` that `SDKControlPermissionRequest` actually documents. They are separate field-mapping sites that drift independently — the same shape as the two hook-input builders.

Adding the field to both turned up something worth flagging: **`handleSDKPermissionRequest` was building `PermissionContext{}` and discarding everything else.** A host served over that path saw no `tool_use_id`, no `agent_id`, no `requires_user_interaction`, no `suppress_always_allow_rule`, no `matched_ask_rule` — every context field added since v0.3.201, silently absent. Only the legacy path was ever populated.

Adding `default_to_no` to an empty struct would have shipped one more field nobody receives, so the context is populated properly here. That means four inbound fields added to `SDKControlRequestBody`, which had carried only `tool_name`, `input`, `tool_use_id` and `agent_id`.

Scope call: this is a bug fix, not a refactor, and it is the smallest change that makes the field this PR is about actually observable. Calling it out rather than burying it in the diff.

## Zero value is the safe one

Plain `bool` throughout. An ask with nothing declared renders as an ordinary ask, which is the correct default for every one of these fields — `false` never causes a host to under-protect. No pointer needed.

## Testing

`TestProtocolPermissionDefaultToNo` covers both paths separately. The SDK-path case asserts the siblings alongside `default_to_no` rather than assuming them, since that context was empty until this change.

`TestIntegrationPermissionRequestContext` provokes a real ask and passes live:

```
ask: tool_use_id="toolu_01N73djTzmtERAFKYbb7u1pm" agent_id="" requires_interaction=false suppress_always_allow=false default_to_no=false
```

Worth noting how it gets there: an in-process SDK MCP server is auto-allowed and never produces an ask at all — the first two drafts skipped for that reason. It takes a stdio server under `WithStrictMCPConfig(true)`, matching `TestIntegrationPermissionCallback`. The assertion is on `tool_use_id` being non-empty, which is the field that is always on the wire, so an unpopulated context fails rather than skips.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3139,3 +3139,75 @@ func TestIntegrationAmbientTaskMarker(t *testing.T) {
 	assert.False(t, userWorkSet,
 		"a subagent the user asked for is not a housekeeping task")
 }
+
+// TestIntegrationPermissionRequestContext provokes a real permission ask and
+// records the context the CLI attached. default_to_no is new in TS SDK
+// v0.3.251; the siblings are asserted because the SDK-format path used to
+// deliver an empty context regardless of what the CLI sent.
+func TestIntegrationPermissionRequestContext(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	// An in-process SDK MCP server is auto-allowed, so the ask has to come
+	// from a stdio server under a strict config — the same setup
+	// TestIntegrationPermissionCallback uses.
+	mcpServerPath := filepath.Join(t.TempDir(), "example-mcp-server")
+	buildCmd := exec.Command("go", "build", "-o", mcpServerPath,
+		"./cmd/example-mcp-server")
+	buildCmd.Dir = "."
+	if out, err := buildCmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to build MCP server: %v\n%s", err, out)
+	}
+
+	var (
+		mu       sync.Mutex
+		captured []PermissionContext
+	)
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("When asked to add numbers you MUST use the "+
+			"add_numbers tool. Do not calculate manually."),
+		WithMCPServers(map[string]MCPServerConfig{
+			"example": {Type: "stdio", Command: mcpServerPath},
+		}),
+		WithStrictMCPConfig(true),
+		WithPermissionMode(PermissionModeDefault),
+		WithCanUseTool(func(
+			ctx context.Context, req ToolPermissionRequest,
+		) PermissionResult {
+			mu.Lock()
+			captured = append(captured, req.Context)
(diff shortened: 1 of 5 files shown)


Roasbeef:
## Review — PR-08, v0.3.251 catchup

Reviewed against `sdk.d.ts` L4055–L4079.

**The empty-context finding is the substance here.** `handleSDKPermissionRequest` constructing `PermissionContext{}` meant every field added to that struct since v0.3.201 — `requires_user_interaction` (v0.3.201), `suppress_always_allow_rule` and `matched_ask_rule` (v0.3.215) — was unreachable for any host served over the SDK-format path, along with `tool_use_id` and `agent_id`. Three prior catchups added fields to a builder that was never wired.

Fixing it inside this PR is the right call rather than splitting it out. `default_to_no` on an empty struct is a field nobody receives, so shipping it alone would have been indistinguishable from doing nothing. The PR body flags it explicitly instead of letting it ride as an unremarked diff hunk, which is what makes it reviewable.

Same dual-site lesson as the hook builders (#173, and PR-01 of this cycle). Two independent field-mapping switches for one wire concept is a standing hazard in this codebase; worth noting it has now bitten in both the hook and permission paths.

**Plain `bool` is correct here** and deliberately different from the `*bool` choices in #221 and #223. Every one of these fields is a *constraint on rendering*, and absent means unconstrained. `false` never causes a host to under-protect, so there is no absent-vs-zero distinction to preserve. The reasoning is stated rather than left as an inconsistency for a reader to wonder about.

**Test discipline.** The SDK-path case asserts the siblings alongside the new field rather than trusting them — appropriate, since they were empty until this change and a test that only checked `DefaultToNo` would have passed against a still-half-wired builder.

The integration test's route to a real ask is the interesting part: in-process SDK MCP servers are auto-allowed and produce no ask, so it takes a stdio server under `WithStrictMCPConfig(true)`. Two drafts skipped before that was pinned down, and the note is in the test so the next person does not repeat it. Passes live with a populated context:

```
tool_use_id="toolu_01N73djTzmtERAFKYbb7u1pm" ... default_to_no=false
```

Asserting on `tool_use_id` rather than on `default_to_no` is right — it is always on the wire, so an unpopulated context fails loudly instead of skipping.

VERDICT: APPROVE

PR #226, opened 2026-08-31
options: price a resume on the SessionStart hook input

Roasbeef:

Four fields on `SessionStartHookInput` (sdk.d.ts L5351–L5367):

```
seconds_since_last_response?  number
context_tokens?               number
prompt_cache_likely_expired?  boolean
estimated_cache_write_usd?    number
```

Resuming a session re-sends its context, and if the prompt cache has aged past its TTL that means paying for a full cache write. These put the numbers in front of the hook that runs at exactly that moment — how old the transcript is, how large a window would be re-sent, whether the cache has likely expired, and what re-caching costs on the session model — so a `SessionStart` hook can decide whether injecting more context is worth it rather than guessing.

Same vocabulary as the `PreModelSwitch` / `PostModelSwitch` inputs from #218. Both surfaces exist so a host can price a cache-invalidating event before committing to it.

All four are scoped to `source: "resume" | "fork"` and are pointer-typed. A plain startup carries none of them, and a zero would read as "free to resume" — the opposite of "unknown". `context_tokens` is also narrower than it sounds: for a server-side tool loop it is the last iteration's window, not the summed totals. Noted in the comment.

## Two dead fields fixed on the way through

Both are in code this change already touches, and both are invisible to a diff-driven catchup:

1. **`SessionStartInput` never modeled `model`.** Upstream has carried it since before this cycle.
2. **`BaseHookInput.PromptID` has been declared since v0.3.201 and neither builder ever populated it.** Every hook input in the SDK has been handing callers an empty prompt-correlation id since the field landed — the struct field existed, the wire key was never read. One line in each builder.

That is the same failure mode as the empty `PermissionContext` in #225: a field added to a type without being wired into the builder that fills it. Two instances in one cycle is worth noting.

## Testing

Both hook-input builders covered separately, plus a startup case asserting all five pointers stay nil — the case that would catch a builder defaulting them to zero.

`TestIntegrationSessionStartResumeCost` builds a real transcript and resumes it. Getting there took working around `isolatedClientOptions`, which sets `WithNoSessionPersistence()` and so makes resume structurally impossible — the first draft hung for the full 150s context. It now shares a config dir with persistence on while still staying out of `~/.claude`. The resume succeeds, but CLI 2.1.222 fires no `SessionStart` hook to the SDK transport at all, so it skips. Filed in `INTEGRATION-FOLLOWUPS.md` with that detail so the next person does not rediscover the persistence trap.


Roasbeef:
## Review — PR-09, v0.3.251 catchup

Reviewed against `sdk.d.ts` L5345–L5367.

**Pointer typing is the right call and for a stated reason.** All four fields populate only on `resume` and `fork`. A zero `estimated_cache_write_usd` reads as "free to resume", which is the exact opposite of "this CLI did not tell you". Same for `context_tokens`. The startup test case asserts all five stay nil, which is what would catch a builder that quietly zero-fills.

**`context_tokens` is narrower than the name.** Upstream: for a server-side tool loop it is the *last iteration's* window, not summed totals. Carried into the comment. A host summing it across a tool loop would overstate the resume cost by a large factor.

**Two dead fields, and this is now a pattern.** `BaseHookInput.PromptID` has been declared since v0.3.201 with neither builder ever reading `prompt_id` — every hook input in the SDK has handed callers an empty correlation id for five cycles. `SessionStartInput.model` was simply never modeled.

Combined with the empty `PermissionContext` found in #225, that is two instances this cycle of *a field added to a struct without wiring the builder that fills it*. The struct compiles, the tests that construct the struct directly pass, and only a test that goes through the builder catches it. Worth carrying forward as a review check: for any inbound shape, confirm the field is read in the builder, not just declared on the type.

Both fixes are in code this diff already touches, so folding them in is correct rather than scope creep.

**Integration test.** The `isolatedClientOptions` finding is genuinely useful and correctly written down: that helper sets `WithNoSessionPersistence()`, so any test built on it cannot resume — the first draft hung for the full 150s context rather than failing usefully. The rewrite shares a config dir with persistence on and stays out of `~/.claude`, and now runs in 6.7s.

The skip is honest: the resume itself succeeds, so the precondition is real; CLI 2.1.222 just fires no `SessionStart` hook to SDK transports. The followup entry records the persistence trap so it is not rediscovered.

Nothing to change.

**CI note.** Lint failed once on `golangci-lint config verify` fetching `golangci-lint.run/jsonschema/...` — a network fetch of a remote schema, unrelated to this diff. Green on re-run. Same flake seen in prior cycles; `golangci-lint run` is clean locally.

VERDICT: APPROVE

Issue #227, opened 2026-08-31
Settings.spinnerTipsOverride.tips: model the object element form


Roasbeef:
Follow-up from #217 (v0.3.251 parity), deferred out of the Settings parity PR.

v0.3.251 widened the element type of `Settings.spinnerTipsOverride.tips`:

```ts
- tips: string[]
+ tips?: (string | { [k: string]: unknown })[]
```

The prose alongside it documents the object form as `{id, text, cooldownSessions?, priority?}` — an entry with a stable id, a cooldown so a tip does not repeat for N sessions, and a priority. The declared type is deliberately open, so the two do not agree on how tight the shape is.

The Go SDK models this as `SettingsSpinnerTipsOverride.Tips []string`. The parity PR took the non-breaking half — `tips` became optional, and `tipsFile` and `label` were added — and left the union alone.

## Why it was deferred

Widening `Tips` to a union element type changes the type of an exported field, which is a breaking change for anyone constructing the struct. Against that: `Settings` only flows *outward* from this SDK — it is authored by the host and handed to the CLI via `--settings` / `--managed-settings` — so there is no decode path that can currently receive the object form and fail on it. The cost of waiting is that a host cannot express cooldowns or priorities, not that anything breaks.

## What it would take

`Tips []SettingsSpinnerTip` with `MarshalJSON` / `UnmarshalJSON` on the element, encoding a bare string as a string and a populated entry as an object. The open index signature means the element also needs an `Extra map[string]any` escape hatch, or the named fields plus passthrough.

## Worth checking first

Whether `Settings` is ever decoded on the resolved-settings transport. `INTEGRATION-FOLLOWUPS.md` refers to fields "round-tripping through the resolved-settings transport", and if that path can hand us a user's managed settings containing an object-form tip, `[]string` would fail the whole `Settings` decode rather than just that field — which would upgrade this from a missing feature to a correctness bug. That question should be settled before deciding the shape.

PR #228, opened 2026-08-31
options: add v0.3.251 Settings parity fields

Roasbeef:

Five scalars plus one reshape:

| field | sdk.d.ts |
|---|---|
| `desktopSessionCleanupPeriodDays` | L5600 |
| `syncClaudeAiPlugins` | L5607 |
| `managedSourcesBehavior` | L7367 |
| `promptCacheTtl` | L7679 |
| `subagentPromptCacheTtl` | L7683 |
| `spinnerTipsOverride` reshape | L7633 |

## CacheTTL is reused, not redeclared

The two TTLs are the same `'5m' | '1h'` set as `cache_ttl` on the model-switch hook inputs from #218. That type is defined once and used in both places. Two independent declarations of the same closed set is how they drift.

## Zero means something here

`desktopSessionCleanupPeriodDays` is `*int`. Zero is the documented **default** and means "no ceiling" — and upstream is explicit that zero is allowed here precisely because it is not allowed on `cleanupPeriodDays`, where it would disable transcript writes. This setting only bounds an *exemption from deletion*; it never stops writes. Collapsing zero into unset would erase the default. `TestSettingsDesktopCleanupZeroSurvives` pins it.

`managedSourcesBehavior` gets the long comment because the merge rules have a security shape: under `"merge"` every present managed source contributes, but the restriction allowlists and the auth pins deliberately do *not* union — the highest source owns them whole. Without that, a lower source could widen a restriction an admin set. Upstream also warns to enable it only when every lower source is admin-controlled, since they start contributing `permissions.allow` entries.

## The spinnerTipsOverride reshape

```
- tips: string[]
+ tips?: (string | { [k: string]: unknown })[]
+ tipsFile?: string
+ label?: string
```

Taken non-breaking. `Tips` keeps `[]string` and gains `omitempty` — required to optional is not a breaking change for callers, and it fixes a real bug: an override that only points at a `tipsFile` was emitting `"tips": []`, which the CLI reads as "no tips". `TestSettingsSpinnerTipsOverrideFileOnly` covers that.

The union element is **not** modeled. Widening an exported field's type is breaking, and `Settings` only flows outward from this SDK, so no decode path can encounter the object form today. Deferred to #227, which also flags the question that should settle it: whether the resolved-settings transport can ever hand us a `Settings` to decode, since that would make it a correctness bug rather than a missing feature.

## Testing

Round-trip and omit-empty pairs matching the existing `TestSettingsParityV0_3_241*` shape, plus the two edge cases above.

The live test was written **after** probing the bare CLI directly:

```
claude --managed-settings '{"desktopSessionCleanupPeriodDays":14,...,"spinnerTipsOverride":{...}}' -p 'Say OK.'
```

That order matters here. An unrecognized member of a managed-settings union stalls the handshake rather than being ignored, which presents as a hang and costs the full timeout — so probing first is cheaper than debugging a stuck integration test. All six keys pass. `TestIntegrationSettingsParityV0_3_251` then pushes them through `WithManagedSettings` and completes a turn, which is the assertable form of "the CLI understood the shape".

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3320,3 +3320,52 @@ func TestIntegrationSessionStartResumeCost(t *testing.T) {
 			"resume-cost fields on the SessionStart hook input")
 	}
 }
+
+// TestIntegrationSettingsParityV0_3_251 pushes the v0.3.251 Settings additions
+// through the managed tier and completes a turn. An unrecognized member of a
+// managed-settings union stalls the handshake outright rather than being
+// ignored, so a completed turn is real evidence the CLI understood the shape —
+// including the reshaped spinnerTipsOverride block.
+func TestIntegrationSettingsParityV0_3_251(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	desktopRetention := 14
+	syncPlugins := false
+	excludeDefault := false
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("You are a helpful assistant. Be very brief."),
+		WithManagedSettings(Settings{
+			DesktopSessionCleanupPeriodDays: &desktopRetention,
+			SyncClaudeAiPlugins:             &syncPlugins,
+			ManagedSourcesBehavior:          "merge",
+			PromptCacheTTL:                  CacheTTL1h,
+			SubagentPromptCacheTTL:          CacheTTL5m,
+			SpinnerTipsOverride: &SettingsSpinnerTipsOverride{
+				ExcludeDefault: &excludeDefault,
+				Tips:           []string{"run make lint before pushing"},
+				Label:          "Ops",
+			},
+		}),
+		WithMaxTurns(1),
+	)
+	client, err := NewClient(opts...)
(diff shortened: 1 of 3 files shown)


Roasbeef:

Verified all six `sdk.d.ts` citations against `scratch/ts-sdk-0.3.251/package/sdk.d.ts` — L5600, L5607, L7367, L7640, L7679, L7683 all land on the claimed declarations and the doc comments compress the upstream prose without inventing semantics.

Two judgment calls worth calling out, both correct:
- `DesktopSessionCleanupPeriodDays` stays a `*int` because zero is meaningful here ("no ceiling") in a way it is not on `CleanupPeriodDays`. `TestSettingsDesktopCleanupZeroSurvives` pins that.
- `Tips` narrowing to `[]string` while upstream widened the element to a union. Settings only flows outward from this SDK, so the union costs us nothing today and widening the field later would be breaking. Deferring is the right call; `omitempty` on `Tips` correctly avoids emitting an empty array the CLI would read as "no tips".

VERDICT: APPROVE

PR #229, opened 2026-09-07
options: curate the model picker and price it at contracted rates

Roasbeef:
Closes the last open surface from the v0.3.251 catchup. See `memory/catchup-v0.3.251/PLAN.md` §"PR-11".

Two nested `Settings` blocks that were split out of the scalar parity PR (#228) because they need their own named structs and their own tests.

- **`modelPicker`** (`sdk.d.ts` v0.3.263 L5921) — an ordered list of `/model` rows an administrator writes, independent of the built-in lineup. `availableModels` still filters the rows, and the highest-precedence source that defines the block wins outright with no merging.
- **`behavesAs`** (L5941) — new in v0.3.263, so the picker rows carry it from the start rather than needing a follow-up. It lends a known model's prompt profile, capability and effort defaults to a model this CLI version does not know. Without it such a row is not offered at all until Claude Code is updated, which makes this the difference between a new model being selectable today and only after a release.
- **`modelPricing`** (L5951) — contracted rates and/or a multiplier, moving every spend figure the CLI reports. This is the setting that makes `ModelUsage.CostBasis` reach `ModelCostBasisManaged` (#220) and the model-switch hooks' `pricing` field reach `configured` (#218); both landed with comments pointing here, and neither value was reachable until now.

**Shape notes:**
- `SettingsModelPricingRates` uses plain `float64`, not pointers. All four rates are required within a row upstream and an invalid row is skipped whole, so eliding a zero would turn a legitimately free model into an incomplete row. `TestSettingsModelPricingZeroRatesSurvive` pins that.
- `SettingsModelPicker.Options` has no `omitempty` for the same reason — `options` is required upstream, and dropping the key would read as "no picker configured" rather than "a picker with no rows".
- `Multiplier` and `ReplaceBuiltInOptions` are pointers: both have meaningful zero values (`0` is an out-of-range multiplier worth sending so the CLI reports it; `false` is the documented default worth stating explicitly).

**Tests:** round-trip and `omitempty` coverage for both blocks, plus the two zero-value pins above. `TestIntegrationSettingsModelPickerPricing` pushes both through the managed tier against the live CLI — an unrecognized member of a managed-settings union stalls the handshake outright rather than being ignored, so a completed turn is real evidence the CLI parsed the nested shapes. Verified passing locally (5.7s).

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3369,3 +3369,63 @@ func TestIntegrationSettingsParityV0_3_251(t *testing.T) {
 	assert.True(t, gotResult,
 		"a stalled managed-settings handshake shows up as no result at all")
 }
+
+// TestIntegrationSettingsModelPickerPricing pushes a curated picker and a
+// contracted price table through the managed tier and completes a turn. Both
+// blocks are nested objects, and an unrecognized member of a managed-settings
+// union stalls the handshake outright rather than being ignored, so a completed
+// turn is evidence the CLI understood the shapes — including behavesAs, which
+// v0.3.263 added inside the picker rows.
+func TestIntegrationSettingsModelPickerPricing(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	replace := false
+	multiplier := 0.85
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPrompt("You are a helpful assistant. Be very brief."),
+		WithManagedSettings(Settings{
+			ModelPicker: &SettingsModelPicker{
+				Options: []SettingsModelPickerOption{{
+					Model:       "opus",
+					Label:       "Deep work",
+					Description: "Architecture and hard bugs",
+					BehavesAs:   "claude-opus-4-8",
+				}},
+				ReplaceBuiltInOptions: &replace,
+			},
+			ModelPricing: &SettingsModelPricing{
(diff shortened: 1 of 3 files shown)


Roasbeef:

Verified `modelPicker` (L5921), `behavesAs` (L5941) and `modelPricing` (L5951) against `scratch/ts-sdk-0.3.263/package/sdk.d.ts`, and cross-checked `behavesAs` against 0.3.251 to confirm it is genuinely new in this version and genuinely a picker-row field — the raw diff makes it easy to misfile onto `availableModels`, which is a plain `string[]`.

The three pointer/value calls are each load-bearing and each pinned by a test:
- `SettingsModelPricingRates` as plain `float64` — a free model prices at zero across the board, and `omitempty` would turn that into a row upstream skips whole.
- `Options` without `omitempty` — a dropped `options` key reads as "no picker", not "an empty picker".
- `Multiplier`/`ReplaceBuiltInOptions` as pointers — both have meaningful zero values.

Probed the bare CLI with `--managed-settings` before trusting the integration assertion, per the standing gotcha that an unknown nested union member stalls the handshake rather than erroring. It parsed clean, and `TestIntegrationSettingsModelPickerPricing` passes against the live CLI in 5.7s.

This closes the last open surface of the v0.3.251 cycle.

VERDICT: APPROVE

Issue #230, opened 2026-09-07
Parity with TypeScript Agent SDK v0.3.263


Roasbeef:
Tracking issue for the weekly catchup against `@anthropic-ai/claude-agent-sdk@0.3.263` (published 2026-09-07). Prior shipped parity: v0.3.251 (#217).

362-line `sdk.d.ts` diff, roughly 70 non-comment lines. Unlike the last two cycles this one carries real transport work: two new CLI flags (`--await-initialize`, `--permission-prompts`) and two new control-protocol requests (`update_settings`, `reload_output_styles`) — the first new control subtypes since `get_settings`. `agentSdkTypes.d.ts` is byte-identical. Nothing was removed and no shape narrowed.

Three themes run through it:

1. **Command-line pressure relief.** `pluginDelivery: 'initialize'` exists because Windows refuses a command line over 32,767 characters and one `--plugin-dir` per plugin blows through it. The list moves to stdin inside the initialize request.
2. **Prompt-batch correlation.** `user_message_uuid` (#222) turns out to be insufficient — when the host merges several close-together sends into one turn it names only the last member. `user_message_uuids` lists every member so any sender can find its own.
3. **Host affordances.** `update_settings`, `reload_output_styles`, `permissionPrompts: 'none'` and the MCP `resource_links` are all things a host driving the SDK needs and could not previously reach.

### Planned PRs

- [x] PR-01 — plugin delivery over initialize (`pluginDelivery`, `--await-initialize`, `plugins`, `plugins_applied`)
- [x] PR-02 — `permissionPrompts` option + `--permission-prompts`
- [x] PR-03 — system prompt `snapshot` + `systemPromptSnapshot`
- [x] PR-04 — `update_settings` control request
- [x] PR-05 — `reload_output_styles` control request
- [x] PR-06 — `detail` on get_context_usage, `skip_behaviors` on get_usage
- [x] PR-07 — `SDKMcpResourceLink` + `resource_links` on task notifications
- [x] PR-08 — `user_message_uuids` batch correlation
- [x] PR-09 — result timing telemetry
- [x] PR-10 — `no_response` on API retry
- [x] PR-11 — `thinkingTokens` on `ModelUsage`
- [x] PR-12 — Settings parity v0.3.263
- [x] PR-13 — docs refresh to v0.3.263 (covers the v0.3.251 tail too)

### Deferred

[...]

PR #231, opened 2026-09-07
transport: actually pass configured plugins to the CLI

Roasbeef:
Prerequisite for the v0.3.263 `pluginDelivery` work (#230, PR-01). Splitting it out because it is a standalone bug fix, not parity work.

`Options.Plugins` has never done anything. `WithPlugins` stored the slice and nothing read it — the transport built no `--plugin-dir` flag, and the initialize request carried no plugin list. Configure a plugin, get a session without it, no error.

This came up while modelling v0.3.263's `pluginDelivery` option, which chooses between argv delivery and delivery over the initialize request. The `'argv'` branch is the default and it did not exist.

**Changes:**
- Emit the flag pair the TS SDK emits (`sdk.mjs` argv builder): `--plugin-dir`, or `--plugin-dir-no-mcp` when `SkipMcpDiscovery` is set. These are alternative flags, not a flag plus an argument, so they cannot be collapsed.
- Preserve configuration order — a later plugin can shadow an earlier one.
- Reject an unsupported `Type` at `Connect` rather than skipping it. The TS SDK throws on the same input. Quietly dropping a plugin the caller asked for is the bug being fixed, so ignoring a type we cannot express would reintroduce it in a narrower form.
- Add `PluginTypeLocal` so callers stop writing the bare string.

**On the strictness:** an empty `Type` now errors too. That is technically a behavior change for code that compiled before, but no such code can have been working — the field was inert, so there is no caller whose plugins were loading and now stop. A clear error at `Connect` beats the previous silence.

**Tests:** flag mapping for both variants, order preservation, no-flag-when-unconfigured, and a table over rejected types asserting the CLI never starts. `TestIntegrationPluginDirArgv` writes a minimal plugin directory and asserts the plugin name appears in the init message's plugin list — asserting only that the session starts would have passed against the broken behavior just as well. Both integration tests verified passing locally (5.3s).

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3429,3 +3429,95 @@ func TestIntegrationSettingsModelPickerPricing(t *testing.T) {
 	assert.True(t, gotResult,
 		"a stalled managed-settings handshake shows up as no result at all")
 }
+
+// writeProbePlugin lays down the smallest plugin directory the CLI will load
+// and returns its path.
+func writeProbePlugin(t *testing.T, name string) string {
+	t.Helper()
+
+	dir := t.TempDir()
+	manifestDir := filepath.Join(dir, ".claude-plugin")
+	require.NoError(t, os.MkdirAll(manifestDir, 0o755))
+
+	manifest := fmt.Sprintf(
+		`{"name":%q,"description":"integration probe","version":"0.0.1"}`,
+		name)
+	require.NoError(t, os.WriteFile(
+		filepath.Join(manifestDir, "plugin.json"), []byte(manifest), 0o644))
+
+	return dir
+}
+
+// TestIntegrationPluginDirArgv asserts that a configured plugin actually
+// reaches the CLI. Options.Plugins was inert before this — the transport built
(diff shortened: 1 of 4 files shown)


Roasbeef:

Confirmed the gap before reviewing the fix: `grep -rn 'plugin-dir' --include=*.go` matched only a comment in options.go, and the initialize request in protocol.go carries no plugin field either. `Options.Plugins` was reachable from `WithPlugins` and read by nothing.

Flag mapping matches the TS argv builder exactly, including the detail that `--plugin-dir-no-mcp` is an alternative flag rather than `--plugin-dir` plus a modifier.

The strictness call is right. Erroring on an unsupported type — empty `Type` included — is a behavior change only in the sense that previously-compiling code now fails; it cannot break a working caller, because no caller's plugins were loading. The failure mode being fixed is silence, so trading it for a narrower silence would miss the point.

Integration coverage is the part worth calling out: asserting the plugin name appears in the init message's plugin list is a test that would have failed against the old behavior. A start-succeeds assertion would have passed against the bug.

VERDICT: APPROVE

PR #232, opened 2026-09-07
options: deliver plugins over the initialize request

Roasbeef:
Parity with TypeScript Agent SDK v0.3.263 — see #230, PR-01. Builds on #231, which made `Options.Plugins` reach the CLI at all.

Adds `Options.PluginDelivery` (`sdk.d.ts` L1889), the `--await-initialize` flag, `plugins` on the initialize request (L4028), and `plugins_applied` on the initialize response (L4033).

**Why it exists:** argv delivery spends one `--plugin-dir` flag per plugin, and Windows refuses to start a process whose command line exceeds 32,767 characters. `PluginDeliveryInitialize` moves the list to stdin so the launch command stops growing with the plugin count. Loading is otherwise identical.

**The exclusivity is the load-bearing part.** `--await-initialize` makes the CLI wait for the initialize request before doing any plugin work, so the flags are not merely redundant next to it — they would load each plugin a second time at a point the request was meant to own. One predicate (`Options.usesInitializePluginDelivery`) drives both the flag and the request field so they cannot disagree in either direction: a CLI told to wait must receive a list, and a list sent to a CLI not told to wait loads nothing.

**Two deliberate edge cases:**
- An empty plugin list stays on the argv path even under initialize delivery. Nothing needs keeping off the command line, and `--await-initialize` would otherwise impose the 2.1.261 version floor for free. This matches the TS gate (`pluginDelivery === 'initialize' && plugins?.length > 0`).
- `plugins` is omitted from the request under argv delivery. The CLI ignores the field without `--await-initialize`, so sending it would claim a delivery that did not happen.

`PluginsApplied` is documented as terminal rather than advisory: the request's list never loads anything after launch, so a `false` means the running process does not have those plugins and will not get them.

**Tests:** transport coverage for flag/no-flag across delivery modes, the empty-list case, invalid-delivery rejection, and no argv leakage under `--await-initialize`; three new cases in the `TestProtocolInitializeOptions` table for the request field's gate.

**Integration:** `TestIntegrationPluginDeliveryInitialize` asserts both that the plugin loads and that `plugins_applied` is true. It is **version-gated, not skipped** — this environment ships CLI 2.1.222, below the 2.1.261 floor (confirmed: `--await-initialize` exits 1 with `unknown option`), so it skips here and starts running on its own once the CLI catches up. `TestCompareDottedVersions` pins the gate's comparator, since a bug there would silently disable the test forever rather than failing.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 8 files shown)


Roasbeef:

Verified the gate against the `sdk.mjs` argv builder rather than the .d.ts alone, since the either/or structure is only visible in the implementation. Upstream computes `jL = pluginDelivery === 'initialize' && plugins !== undefined && plugins.length > 0`, then branches `if (jL) push('--await-initialize'); else if (plugins?.length) <per-plugin flags>`, and separately sets `plugins: jL ? plugins : undefined` on the init config. Deriving both sites from one predicate reproduces that exactly and removes the possibility of the two drifting apart.

The empty-list carve-out is the subtle one and it is right. Upstream's `jL` includes the length check, so an empty list under initialize delivery falls through to the argv branch and emits nothing — no `--await-initialize`. Emitting it anyway would have imposed the 2.1.261 floor on sessions with no plugins at all, which is a real regression for callers who set the option globally.

Confirmed the version floor empirically: this environment's CLI is 2.1.222 and `claude --await-initialize -p` exits 1 with `error: unknown option '--await-initialize'` — exactly the failure the docs describe, and the reason argv has to stay the default.

Version-gating the integration test instead of `t.Skip`-ing it unconditionally is the right call, and pinning the comparator is what makes the gate trustworthy — the failure mode of a broken version compare is a test that never runs again and never complains.

VERDICT: APPROVE

PR #233, opened 2026-09-07
options: let a session declare it has no approval surface

Roasbeef:
Parity with TypeScript Agent SDK v0.3.263 — see #230, PR-02.

Adds `Options.PermissionPrompts` (`sdk.d.ts` L1836) and the `--permission-prompts` CLI flag.

**The gap it closes:** an unattended session has nobody to ask. Before this the SDK could only let a call that would prompt sit waiting for an answer that never arrives, or pre-approve tools it did not actually want to pre-approve. `PermissionPromptsNone` gives a third option — deny immediately and tell Claude the session has no approval surface, so it can adapt instead of stalling.

**It does not loosen anything.** The permission mode (auto mode's classifier included), the rules and the hooks all still decide exactly as before. Only the fallback changes.

**Three things worth reviewing:**
- **The flag is sent even when `CanUseTool` is registered.** `none` is precisely what tells the CLI to stop short of the prompt tool, so suppressing it in that case would leave the callback wired up and reintroduce the hang this is meant to remove. Upstream is the same shape — it pushes `--permission-prompt-tool stdio` and `--permission-prompts` independently. Setting both is a legitimate pattern: a host installs a callback globally and suppresses prompting for one session.
- **Documented apart from `PermissionMode`.** `PermissionModeDontAsk` sounds like the same feature. It is not: the mode participates in the decision (deny what is not pre-approved), while this says whether an approval surface exists at all, under whichever mode is in force. The type comment spells that out because the names invite the confusion.
- **Invalid values are rejected at `Connect`** rather than passed through to the CLI.

**Tests:** flag mapping for both values, no flag when unset (host is the CLI's own default — the SDK should not start pinning it), the none-plus-callback case, and invalid-value rejection.

**Integration:** `TestIntegrationPermissionPromptsNone` asks for an un-pre-approved Bash call and asserts the callback never fires *and* that `result.permission_denials` is non-empty — a denial actually booked, not merely a session that did not hang.

**On the skip helper:** this environment's CLI (2.1.222) does not know `--permission-prompts`, so the test skips. My first version detected that by waiting for the context to expire, which cost 90 seconds per CI run. `skipIfCLILacksFlag` instead probes with a deliberately invalid value: both outcomes fail during option parsing before any API call, and the error text discriminates (`unknown option` vs an argument complaint). Same skip in 0.19s, and it self-backfills when the CLI updates. The helper is reusable for future catchups, where a flag routinely lands upstream before the bundled CLI catches up.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -3575,6 +3576,30 @@ func compareDottedVersions(a, b string) int {
 	return 0
 }
 
+// skipIfCLILacksFlag skips when the installed CLI does not recognize a flag.
+//
+// The probe passes a deliberately invalid value, which makes both outcomes
+// fail during option parsing — before any API call — and lets the error text
+// tell them apart: an unrecognized flag reports "unknown option", while a
+// recognized one complains about the argument. Cheaper and more precise than
+// letting the real test hang until its context expires.
+func skipIfCLILacksFlag(t *testing.T, flag string) {
+	t.Helper()
+
+	path, err := DiscoverCLIPath(&Options{})
+	if err != nil {
+		t.Skip("claude CLI not found in PATH")
+	}
+
+	cmd := exec.Command(path, flag, "__probe__", "-p", "x")
+	cmd.Stdin = strings.NewReader("")
+	out, _ := cmd.CombinedOutput()
+
+	if strings.Contains(string(out), "unknown option '"+flag+"'") {
+		t.Skipf("CLI does not support %s", flag)
+	}
+}
+
 // awaitInitializeMinCLI is the first Claude Code release that understands
 // --await-initialize. An older binary exits at startup with an unknown-option
 // error (sdk.d.ts v0.3.263 L1885).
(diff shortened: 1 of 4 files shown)


Roasbeef:

Verified against `sdk.d.ts` L1836 and the `sdk.mjs` argv builder. Upstream pushes `--permission-prompt-tool stdio` (when a callback exists) and `--permission-prompts` as independent statements with no mutual exclusion, which is what this reproduces. Keeping both is not merely matching upstream for its own sake — dropping the flag when a callback is present would defeat the entire feature, since the callback is exactly what 'none' is meant to bypass.

Separating this from `PermissionMode` in the docs is the right instinct. `PermissionModeDontAsk` reads like a synonym and is not one, and someone reaching for 'no prompts' will hit that constant first.

The integration assertion earns its keep by checking `permission_denials` rather than just absence of a hang — a session that quietly allowed the call would pass the weaker check.

The skip helper is the most valuable part of this PR beyond the feature. A 90-second-per-run skip is a slow leak in CI time that nobody would have attributed to a skipped test, and probing with an invalid value to force a parse-time failure is a clean way to ask 'does this binary know this flag' without spending an API call. This will get reused — flags land upstream before the bundled CLI catches up on most cycles.

CI green; all checks success or skipped.

VERDICT: APPROVE

PR #234, opened 2026-09-07
protocol: send the preset system prompt's append

Roasbeef:
Prerequisite for the v0.3.263 `systemPrompt.snapshot` work (#230, PR-03). Splitting it out because it is a standalone bug fix, not parity work — same situation as #231.

`WithSystemPromptPreset` stored a `SystemPromptConfig` that no wire path read. `Options.SystemPromptPreset` was matched by nothing in `transport.go` or `protocol.go`, so a caller asking for Claude Code's prompt plus their own instructions got the prompt without the instructions, and no error.

**Changes:**
- Populate `appendSystemPrompt` on the initialize request from the preset, matching the TS derivation (`sdk.mjs`: for `type: 'preset'` it sets only `appendSystemPrompt`, `excludeDynamicSections` and `snapshot`).
- Leave `systemPrompt` unset for the preset form. That absence is what selects the built-in prompt; filling it with the append text would replace the whole preset with a fragment.
- Refuse the combination of a custom `SystemPrompt` and a `SystemPromptPreset`.
- Document that `SystemPromptConfig.Preset` is not sent — `claude_code` is the only value, and the shape of the request is already unambiguous. Upstream reads `s.append` / `s.excludeDynamicSections` / `s.snapshot` off the preset object and never `s.preset`.

**On refusing the conflict:** upstream models this as a union, so the case cannot arise there. As two Go fields it can, and the two are genuinely contradictory — a custom prompt *replaces* the preset rather than extending it, so there is no defensible way to honor both. Erroring beats picking one silently.

**Tests:** three cases added to the `TestProtocolInitializeOptions` table (preset sends only the append; preset with no append sends neither key; custom prompt sends no append) plus a conflict test.

**Integration:** `TestIntegrationSystemPromptPresetAppend` asserts the append actually reaches the model.

Worth flagging how that test got written, since it bears on trusting it: the first version appended *"end every reply with ZORBLAX"* and prompted *"say hello in exactly three words"*. Those two instructions contradict each other, and the model honored one or the other roughly half the time — it failed, passed on a re-run, then failed again. That is a test that would have landed and then flaked in CI for someone else to debug. The version here puts a codeword in the append and asks for it, so compliance is recall rather than obeying a formatting rule that fights the prompt. Ran it three times in a row clean, and it is faster (4-5s vs 7-9s).

I also verified the underlying mechanism independently before trusting either result — a hand-rolled `stream-json` session that sends `appendSystemPrompt` on the initialize request and nothing else confirms the CLI honors it from the control protocol, with no `--append-system-prompt` flag involved (upstream emits no such flag either).

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3747,3 +3747,58 @@ func TestIntegrationPermissionPromptsNone(t *testing.T) {
 		"a tool call that would have prompted must be denied outright, not "+
 			"silently allowed; result: %s", result.Result)
 }
+
+// TestIntegrationSystemPromptPresetAppend asserts that the preset form's
+// append actually reaches the model. WithSystemPromptPreset stored a config no
+// wire path read, so the append was silently dropped; a test that only checked
+// the session ran would have passed against that.
+//
+// The append carries a fact the model cannot otherwise know, and the prompt
+// asks for it. Recalling a fact is far more reliable than obeying a formatting
+// rule: an earlier version of this test appended "end every reply with X" and
+// asked for a three-word greeting, and the model split its compliance between
+// the two contradictory constraints about half the time.
+func TestIntegrationSystemPromptPresetAppend(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	const marker = "ZORBLAX7"
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPromptPreset("claude_code", fmt.Sprintf(
+			"The deployment codeword for this session is %s. "+
+				"When asked for the deployment codeword, reply with it.",
+			marker)),
+		WithMaxTurns(1),
+	)
+	client, err := NewClient(opts...)
+	require.NoError(t, err)
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
+	defer cancel()
+
+	var reply strings.Builder
+	var gotResult bool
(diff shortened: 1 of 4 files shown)


Roasbeef:

Confirmed the gap independently: `grep -rn SystemPromptPreset` matches only its own declaration and setter — no transport or protocol read. Second inert option found this cycle, after `Options.Plugins` in #231.

Checked the derivation against `sdk.mjs` rather than the .d.ts, since the preset/custom asymmetry only shows up in the implementation. Upstream: `if (s.type === 'custom') { p = s.prompt; h = s.snapshot } else if (s.type === 'preset') { f = s.append; m = s.excludeDynamicSections; h = s.snapshot }` — `p` (systemPrompt) is deliberately left undefined for the preset branch, and `s.preset` is never read. Both details are reproduced here, and the second is documented rather than silently dropped, which is the right call for a field that looks load-bearing and is not.

Refusing the custom+preset combination is correct. It is unrepresentable upstream and genuinely contradictory here, so there is no behavior to be compatible with.

The flake writeup is the most useful thing in this PR. A self-contradicting integration assertion that passes on the second run is exactly the kind of test that lands green and then costs someone an afternoon three weeks later. Rewriting compliance-as-recall instead of compliance-as-formatting is the right fix, and three clean consecutive runs plus the independent raw-protocol verification is enough to trust it.

CI green; all checks success or skipped.

VERDICT: APPROVE

PR #235, opened 2026-09-07
options: opt in to system prompt snapshotting

Roasbeef:
Second half of the v0.3.263 system-prompt work (PR-03; #234 landed the preset append).

`snapshot` on the TS SDK's system-prompt config records the conversation's system prompt once and replays it verbatim on every later request and resume. A prompt that shifts mid-conversation invalidates the API prompt-cache prefix and, under extended thinking, discards the model's earlier reasoning — a CLI upgrade between launches, a flipped flag, or a different append is enough to cause it.

## Shape

- `Options.SystemPromptSnapshot *bool` + `WithSystemPromptSnapshot(bool)`.
- Wired to `systemPromptSnapshot` on the initialize control request (top-level, matching how the SDK already flattens `systemPrompt`/`appendSystemPrompt`).

The field is a pointer because the three states are genuinely distinct — omitted is not `false`:

| state | behavior |
|-------|----------|
| `nil` | record only a bare preset; a custom prompt or append renders fresh each launch (historical behavior) |
| `true` | reuse an existing record; a later launch's divergent prompt/append is ignored until compaction |
| `false` | never record; render fresh every request |

## Tests

- Unit: snapshot true/false/unset wiring into the initialize request, plus snapshot riding alongside a preset append (`protocol_test.go`).
- Integration: `TestIntegrationSystemPromptSnapshot` round-trips snapshot=true against the live CLI, asserting the init handshake accepts the field and a paired preset append still reaches the model (passes, 3.6s).

Safe to set before the feature reaches an account: where recording is not yet enabled, and on Bedrock/Vertex/Foundry today, the field is accepted and does nothing (sdk.d.ts v0.3.263 L2219).

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3802,3 +3802,56 @@ func TestIntegrationSystemPromptPresetAppend(t *testing.T) {
 		"the preset append never reached the model; reply was %q",
 		reply.String())
 }
+
+// TestIntegrationSystemPromptSnapshot asserts the CLI accepts a snapshotted
+// system prompt in the initialize handshake and still runs the turn. The
+// snapshot's real effect — a prompt cached once and replayed verbatim across
+// requests — is not observable from a single query, so this guards the wire
+// contract instead: an unknown field in the initialize request stalls the
+// handshake for the full timeout, and pairing snapshot with a preset append
+// confirms the append still lands while snapshotting is on.
+func TestIntegrationSystemPromptSnapshot(t *testing.T) {
+	skipIfNoToken(t)
+	skipIfNoCLI(t)
+
+	const marker = "QUASAR3"
+
+	opts := append(isolatedClientOptions(t),
+		WithSystemPromptPreset("claude_code", fmt.Sprintf(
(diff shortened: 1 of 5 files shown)


Roasbeef:
**VERDICT: APPROVE**

Reviewed against `sdk.d.ts` v0.3.263 L2192–2240 and the `sdk.mjs` wire mapping (`systemPromptSnapshot:this.initConfig?.systemPromptSnapshot`).

- Pointer tri-state (nil/true/false) matches the upstream `snapshot?: boolean` semantics: omitted records a bare preset, `true` reuses, `false` never records. Correct not to elide `false`.
- Flattened to a top-level `systemPromptSnapshot` on the initialize request, consistent with how the SDK already flattens `systemPrompt`/`appendSystemPrompt`.
- Unit tests cover all three states plus the preset-append pairing; integration test round-trips against the live CLI (passed, 3.6s) and guards the init handshake against an unknown-field stall.

CI green (Lint, Test, all Build matrix; Integration skipping as expected in CI). Merging.

PR #236, opened 2026-09-07
client: add update_settings control request

Roasbeef:
PR-04 of the v0.3.263 catchup. Adds the `update_settings` control request (sdk.d.ts L4473, `updateSettings` client method L2708).

`ApplyFlagSettings` merges into the in-memory flag layer and vanishes when the session ends. A host application driving the SDK that wants a setting to survive needs it written to disk — where `/config`'s writes land. `update_settings` persists the merge to a named settings file.

## Shape

- `Stream.UpdateSettings(ctx, source, settings)`.
- Wire: `{subtype:"update_settings", source, settings}` on the control request (confirmed against `sdk.mjs`).
- `source` is a single-member union upstream (`'localSettings'`), modelled as a named `SettingsSource` type with `SettingsSourceLocal` so widening it stays additive.

## CLI contract (documented on the method)

Enforced CLI-side, surfaced as an error response: `outputStyle` is the only writable key today, values must be strings, deletion is unsupported, and it refuses remote transports and any session whose `--setting-sources` excludes the target.

## Tests

- Unit: source + settings wire shape, and the empty-map (nil) call sending `settings:{}` with `source` still present (`client_file_plugin_control_test.go`).
- Integration: `update_settings` subtest under `TestIntegrationStreamFileAndRuntime` writes `outputStyle` and reads it back off disk to prove the write landed. Guarded with `skipIfCLIOlderThan("2.1.261")` — the subtype lands in Claude Code 2.1.261 and the CI/local binary (2.1.222) rejects it as unsupported. Skips cleanly today; runs once the runner updates.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1196,6 +1196,40 @@ func (s *Stream) ApplyFlagSettings(
 	return err
 }
 
+// SettingsSource names the settings file Stream.UpdateSettings writes to. It
+// is a single-member type today; a named type keeps widening it additive.
+type SettingsSource string
+
+// SettingsSourceLocal is the project's local settings file — the only scope the
+// CLI accepts today, chosen so a host UI's writes land exactly where /config's
+// do.
+const SettingsSourceLocal SettingsSource = "localSettings"
+
+// UpdateSettings writes settings to a settings file on disk, the durable
+// counterpart to ApplyFlagSettings' in-memory flag layer. The CLI persists the
+// merge to the named file rather than holding it for the session.
+//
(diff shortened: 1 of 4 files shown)


Roasbeef:
**VERDICT: APPROVE** — Wire shape `{subtype:"update_settings",source,settings}` matches sdk.mjs; `source` modelled as a widening-safe named type; CLI contract documented. Unit tests assert the wire shape; integration test reads the persisted `outputStyle` back off disk, guarded on 2.1.261. CI green (Lint, Test, Build matrix). Merging.

PR #238, opened 2026-09-07
client: add options to get_context_usage and get_usage

Roasbeef:
PR-06 of the v0.3.263 catchup. Adds two independent options to existing control requests, grouped because the edit shape is identical (sdk.d.ts L3627 `detail`, get_usage `skip_behaviors`).

Both methods take an options struct via a variadic parameter, so existing no-arg callers keep compiling.

## get_context_usage — `detail`

`ContextUsageDetail` (`full`|`summary`). `summary` answers from the last response's usage and local estimates, skipping the per-category token-count API calls; `full` is the CLI default when the field is omitted. Empty value defers to the CLI.

## get_usage — `skip_behaviors`

`GetUsageOptions.SkipBehaviors` nulls the response's `behaviors` section. The scan that fills it reads every transcript touched in the last seven days, so a caller that needs only the plan rate limits (a usage meter) can drop it. Sent only when true (matches `sdk.mjs`, which spreads `skip_behaviors:!0` conditionally).

## Tests

- Unit: wire-shape assertions for both — `detail:"summary"` present / omitted when unset, `skip_behaviors:true` present / omitted when false (`client_introspection_test.go`).
- Integration: `TestIntegrationGetContextUsageDetail` runs `detail=summary` against the live CLI and asserts a token total comes back (passes, 3.9s; the 2.1.222 runner ignores the unknown field and still answers). `TestIntegrationGetUsageSkipBehaviors` is a documented skip — `get_usage` is EXPERIMENTAL and account-dependent (mirrors the existing `TestIntegrationGetUsageExperimental` skip); the option's wire shape is unit-covered.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1432,12 +1432,37 @@ func (s *Stream) AccountInfo(ctx context.Context) (*AccountInfo, error) {
 	return &account, nil
 }
 
+// ContextUsageDetail selects how GetContextUsage computes its answer.
+type ContextUsageDetail string
+
+const (
+	// ContextUsageDetailFull counts each category with the token-count API.
+	// This is the CLI's default when the field is omitted.
+	ContextUsageDetailFull ContextUsageDetail = "full"
+
+	// ContextUsageDetailSummary answers from the last response's usage and
+	// local estimates, skipping the per-category token-count API calls.
+	ContextUsageDetailSummary ContextUsageDetail = "summary"
+)
+
+// GetContextUsageOptions configures Stream.GetContextUsage.
+type GetContextUsageOptions struct {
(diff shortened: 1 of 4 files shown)


Roasbeef:
**VERDICT: APPROVE** — `detail` and `skip_behaviors` match sdk.d.ts/sdk.mjs; variadic options keep no-arg callers compiling; `skip_behaviors` sent only when true (matches conditional spread). Unit tests assert present/omitted wire shapes; `detail` integration runs live, `skip_behaviors` skip mirrors the experimental get_usage base test. CI green. Merging.

PR #239, opened 2026-09-07
messages: add resource_links to task_notification

Roasbeef:
PR-07 of the v0.3.263 catchup. Adds `SDKMcpResourceLink` (sdk.d.ts L4629) and `resource_links` on `TaskNotificationMessage` (L5225).

A backgrounded MCP task (`task_type: mcp_task`) reports a placeholder text `tool_result`; its real result arrives asynchronously on the `task_notification`. `resource_links` carries the `resource_link` content blocks of that result — the files the tool returned by reference — which is the only place a host learns what the call produced. Join to the originating call via `tool_use_id`.

## Shape

```go
type SDKMcpResourceLink struct {
    URI, Name          string
    Title, Description  string  // optional
    MimeType            string  // optional
    Size               *int     // optional
    Annotations         map[string]interface{}
}
```

`Size` is a pointer so a zero-byte link is distinguishable from one that omits the field.

## Notes (documented on the field)

Same caps as the inline tool result — at most 50 links, 64 KiB serialized. Absent when the result had none or the task is not an `mcp_task`, and never populated from the server's `_meta`.

## Tests

- Unit: a `mcp_task with resource_links` case in `TestParseMessageTaskNotification` parses two links (one full, one zero-byte to prove the pointer), asserts every field, and the `failed` case asserts `ResourceLinks` is nil when absent.
- Integration: `TestIntegrationMcpResourceLinks` is a documented skip — populating it needs a backgrounded `mcp_task` returning `resource_link` blocks, which the standard run can't stand up. Recorded in INTEGRATION-FOLLOWUPS.md; parse path is unit-covered.


Roasbeef:
Reviewed against \`scratch/ts-sdk-0.3.263/package/sdk.d.ts\`.

\`SDKMcpResourceLink\` (L4629) and \`resource_links\` on \`SDKTaskNotificationMessage\` (L5225) match field-for-field, including optionality. The struct sits in TS declaration order, which keeps future diffs against upstream cheap.

Points that earn their keep:

- \`Size *int\` — a link reporting a zero-byte file has to stay distinguishable from one omitting the field. The unit test pins both halves of that, which is the only reason the pointer survives review.
- The doc comment carries the parts of the wire contract a Go caller cannot infer from the type: placeholder \`tool_result\` for backgrounded tasks, the \`ToolUseID\` join, the 50-link / 64 KiB caps, and the never-from-\`_meta\` rule.
- \`TestIntegrationMcpResourceLinks\` skips rather than pretending. Populating it needs a backgrounded \`mcp_task\` fronting an MCP server that returns files by reference — not something the standard run can stand up — and the followup entry says exactly what would have to change. Parse path is covered by unit test.

Non-blocking: \`Mcp\` vs \`MCP\` in the type name. The repo is already split (\`McpServer\`, \`McpServerInfo\` next to \`SDKMCPServerConfig\`, \`MCPServerConfig\`), and mirroring the TS spelling is the more useful tiebreak for a catchup surface. Not worth a rename.

Build/Lint/Test green on all five matrix entries; Integration skipping as expected without a token.

VERDICT: APPROVE

PR #240, opened 2026-09-07
messages: add user_message_uuids to reply, result and thinking frames

Roasbeef:
A host that sends several user messages close together gets back fewer turns than it sent. The host merges the batch into one turn, and that turn's `user_message_uuid` names only the last member — every other sender in the batch has no way to recognize the reply as its own.

TS SDK v0.3.263 fixes that by carrying the whole consumption-ordered list beside the singular field. This models it on the three frames that carry it, and picks up the `thinking_tokens` attribution field while we are in here.

## What the list means

It means something slightly different per site, so the comments say so rather than sharing one blurb:

- **`AssistantMessage` / `PartialAssistantMessage`** — what the turn has consumed *so far*: the merged batch, nothing else. Present exactly when the singular field is, on the same first frame.
- **`ResultMessage`** — everything the turn consumed, *including* queued messages folded in between tool rounds after the reply frame was stamped. That folding is why it can be longer than the list the caller already saw on the reply. Upstream carries it on both `SDKResultSuccess` (L5014) and `SDKResultError` (L4991); Go models those as one struct.
- **`ThinkingTokensMessage`** — only the singular uuid, but on *every* frame rather than just the first, so a consumer can attribute thinking progress to its send before any reply frame arrives (L5327).

Invariants in all cases: the list always contains the singular field, holds at most 64 entries, and is absent from older producers — where the singular field stays the fallback.

## Testing

Unit tests cover the parse path on all four frames, including the longer-than-the-reply result list and the older-producer shape that omits it.

`TestIntegrationUserMessageUUIDs` sends a real turn with a client uuid and asserts the result echoes it, then skips rather than asserting on an absent list: CLI 2.1.222 populates the singular field only. Entry added to `INTEGRATION-FOLLOWUPS.md` describing what has to change to backfill the assertion.


Roasbeef:
Reviewed against `scratch/ts-sdk-0.3.263/package/sdk.d.ts`. All four sites check out:

- `SDKAssistantMessage.user_message_uuids` L3336
- `SDKPartialAssistantMessage.user_message_uuids` L4843
- `SDKResultError` L4991 / `SDKResultSuccess` L5014 — one Go struct covers both, consistent with how the singular field was landed in #222
- `SDKThinkingTokensMessage.user_message_uuid` L5327

The per-site comments are the part worth having. A single shared blurb would have flattened the one distinction a caller actually has to reason about: the reply-frame list is the merged batch, the result list additionally folds in queued messages taken off the queue between tool rounds, so the result list can be strictly longer than the one the caller already saw. `TestParseMessageUserMessageUUIDs/result folds queued messages so the list is longer` pins exactly that, and the older-producer subtest pins the fallback.

`ThinkingTokensMessage` correctly takes only the singular field — upstream gives it no list — and the comment records the two things that make it different from the reply-frame stamp: every frame rather than the first, and absent on Remote Control sessions.

Integration test sends a real turn and asserts the echo before skipping, so it degrades to a live check of the singular field rather than a no-op; followup entry says what has to change to backfill the list assertion.

Two `sdk.d.ts` citations were off in the working copy and were corrected before commit: the thinking-tokens comment pointed at L4839 (`SDKPartialAssistantMessage`, wrong type) and the result comment at a doc-comment line rather than the two field lines.

Build/Lint/Test green on all five matrix entries; Integration skipping without a token.

VERDICT: APPROVE

PR #241, opened 2026-09-07
messages: add the first-frame result timings

Roasbeef:
TS SDK v0.3.263 adds four timing readings to the success result — `first_content_frame_ms`, `first_stream_post_ms`, `first_stream_post_ack_ms`, `first_stream_post_wall_ms` (sdk.d.ts L5016–L5019). They sit beside `request_sent_wall_ms` from v0.3.220 and, like it, are declared on `SDKResultSuccess` only.

## On the comments

Upstream ships no doc comment for any of the four, and `sdk.mjs` is a 225-line wrapper with no computation in it — the type declaration is the entire specification available. So the comments stay at the terse trailing style `request_sent_wall_ms` already uses and claim only what is knowable: the scale each reading is on and that they are success-only. Inventing plausible-sounding semantics here would be worse than saying less, because the next reader cannot tell an inference from a fact.

All four are pointers. A genuine 0ms reading has to stay distinguishable from a producer that never reported one, and the unit test pins a zero `first_content_frame_ms` for that reason.

## Testing

Unit tests cover the parse path, the zero-versus-absent distinction, the `omitempty` marshalling, and the error-result shape where all four are nil.

`TestIntegrationResultFirstFrameTimings` runs a real streaming turn and asserts a result came back before skipping. Probed against CLI 2.1.222: the result carries `ttft_ms`, `ttft_stream_ms`, `time_to_request_ms` and the `duration_*` pair, but none of these four and no `time_origin_ms`. Verified locally — the test skips after 4.95s, so it is exercising a live turn rather than short-circuiting.

Where the CLI does populate them, the test asserts the one ordering that holds by construction (an ack cannot precede the post it acknowledges) and *logs* the origin/relative/wall triple rather than asserting `wall == origin + relative`. That relationship looks obvious but upstream never states it, and a wrong assertion here would fire on whoever backfills the test rather than on me.

## Incidental finding

While probing, CLI 2.1.222 does populate `ttft_stream_ms` and `time_to_request_ms` — which the existing `TestIntegrationResultMessageTimingFields` skip reason says are not triggerable. That skip is stale for two of its four fields. Recorded in `INTEGRATION-FOLLOWUPS.md`; not fixed here, since making it assert would be its own change.


Roasbeef:
Self-review against `scratch/ts-sdk-0.3.263/package/sdk.d.ts` L5016–L5019. Four fields, right names, right JSON tags, placed in declaration order between `request_sent_wall_ms` and `time_to_request_from_spawn_ms` so future diffs against upstream stay cheap.

The judgment call worth recording is what the comments *do not* say. Upstream ships no doc comment for any of the four and `sdk.mjs` carries no computation, so anything beyond scale-and-variant would be invention. The terse trailing style matches what `request_sent_wall_ms` already uses, and a reader can tell fact from inference — which is the whole point.

Same reasoning drove dropping an assertion from the integration test. `wall == origin + relative` reads as obviously true, but upstream never states it, and encoding a guess as an assertion means it fires on whoever backfills the test rather than on the person who guessed. It logs the triple instead.

Pointer choice is load-bearing and tested: `first_content_frame_ms: 0` parses to a non-nil zero, and the error-result case parses to nil for all four.

Integration test verified live, not just compiled — skips after 4.95s, so it genuinely drove a turn and inspected a real result before bailing. CLI 2.1.222 reports none of the four.

VERDICT: APPROVE

PR #242, opened 2026-09-07
messages: add no_response to api_retry

Roasbeef:
TS SDK v0.3.263 adds `no_response?: { waited_ms, retry_wait_ms }` to `SDKAPIRetryMessage` (sdk.d.ts L3307).

The interesting part is not the two durations — it is what the field's *presence* tells you. A retry after a first-byte timeout counts against a cap of its own, normally a single retry, rather than the session's retry budget. From `attempt` and `max_retries` alone a consumer tracking retries against a budget cannot tell the two situations apart, and would mix a one-shot cap into a running count. That is the reason upstream distinguishes the cause at all, so the comment leads with it rather than with the durations.

Modelled as a pointer to a named `APIRetryNoResponse`. Nil is meaningful: it says this is an ordinary retry and `MaxRetries` is the session budget.

## Testing

Extended the existing `TestParseMessageAPIRetry` table with a first-byte-timeout case rather than adding a parallel test, which gets the absent-case assertion for free: both pre-existing rows now assert `NoResponse` is nil, pinning the distinction in both directions.

`TestIntegrationAPIRetryNoResponse` skips. The field populates only when the API accepts a request and then withholds response headers past `CLAUDE_STREAM_FIRST_BYTE_TIMEOUT_MS`, and that timeout is enforced inside the CLI's own HTTP client — the SDK has no way to inject it. Provoking it would need a proxy sitting between the CLI and the API that accepts and stalls. The followup entry says exactly that, so whoever picks it up knows the shape of the fixture required rather than rediscovering why it is hard.


Roasbeef:
Self-review against `scratch/ts-sdk-0.3.263/package/sdk.d.ts` L3307. Shape matches: optional object, two required number members, sitting after `error` and before `uuid` in declaration order.

The modelling call I want on record: `*APIRetryNoResponse` rather than flattening `waited_ms`/`retry_wait_ms` onto `APIRetryMessage` as two more optional scalars. Flattening would lose the thing that actually matters — nil-versus-set on the whole object is the signal that reclassifies `MaxRetries` from session budget to one-shot cap. Two independent optional scalars would let a caller check one and not the other, and would not read as a single condition.

Comment leads with that reclassification rather than with the durations, since the durations are self-describing and the budget semantics are not.

Extending the existing `TestParseMessageAPIRetry` table instead of writing a parallel test earns the absent-case coverage for free: both pre-existing rows now assert nil, so the distinction is pinned in both directions rather than only on the happy path.

The integration test is an honest skip. The first-byte timeout is enforced inside the CLI's HTTP client, so the SDK cannot inject it — a fixture would need a proxy that accepts the request and withholds headers. The followup entry names that fixture so the next person does not re-derive why it is hard.

VERDICT: APPROVE

PR #243, opened 2026-09-07
messages: add thinkingTokens to ModelUsage

Roasbeef:
TS SDK v0.3.263 adds `thinkingTokens?: number` to `ModelUsage` (sdk.d.ts L1313), breaking out the thinking portion of a model's output tokens.

Two traps here, both of which the comment calls out because neither is visible from the type:

1. **It is contained in `OutputTokens`, not additional to it.** A consumer that sums the two double-counts. The unit test asserts `thinking <= output` and the integration test asserts the same invariant against any real entry that carries the field.
2. **It is a partial tally across CLI versions.** It counts only turns that ran on a CLI recording the field, so a session resumed from an older version reports thinking for some turns and not others. That makes it unsafe as a denominator against `OutputTokens` — the ratio would be wrong in a way that looks plausible.

`*int` rather than `int`, because absent and zero are different claims: nil means no turn in this session reported thinking tokens at all, zero means a model that does report them used none. Given trap 2, collapsing those would be the difference between "we cannot tell" and "there was no thinking."

## Testing

Unit tests cover all four states — present and contained, absent, explicit zero, and omitted from the wire when nil.

`TestIntegrationModelUsageThinkingTokens` drives a real turn with a prompt that asks for step-by-step reasoning. Probed against CLI 2.1.222: the result prices two entries (haiku and opus) and neither carries the field, so the test asserts containment for any entry that does and skips otherwise. Verified live — skips after 9.79s, so it is genuinely driving a turn and walking real usage entries.


Roasbeef:
Self-review against `scratch/ts-sdk-0.3.263/package/sdk.d.ts` L1313.

`*int` is the call worth defending. Every other token count on `ModelUsage` is a plain `int`, so this breaks the local pattern — but upstream types it optional for a reason the doc comment states: it counts only turns that ran on a CLI recording the field. That makes nil ("no turn in this session reported thinking") and zero ("a reporting model used none") genuinely different claims. Collapsing them to `int` would report confident zeros for sessions that simply have no data, which is worse than the inconsistency.

The same fact makes it unsafe as a denominator — a session resumed across CLI versions gets a partial numerator against a complete `OutputTokens`, and the resulting ratio looks plausible while being wrong. Comment says so explicitly rather than leaving it to be rediscovered.

Containment (`thinking <= output`, not a sibling to sum) is stated upstream, so it is asserted in both the unit test and the integration test rather than merely documented.

Four unit subtests cover present/absent/explicit-zero/omitted. Integration test verified live — skips after 9.79s having priced two real entries (haiku and opus), neither carrying the field on CLI 2.1.222.

VERDICT: APPROVE

PR #244, opened 2026-09-07
options: settings parity with TypeScript SDK v0.3.263

Roasbeef:
Six `Settings` adds from TS SDK v0.3.263. Most are ordinary scalars; three carry a trap worth commenting.

## The traps

**`blockReadsOutsideWorkingDirectories` nests in the permissions block** (sdk.d.ts L5889), not at the top level. Filed wrong it would decode silently into nothing and the test would still pass, so the unit test asserts both directions: present under `permissions`, absent at top level.

**`timeFormat` is a union, not an enum** (L8076): the four presets *or* any string containing `%`, read as a strftime pattern. Modelling it as a closed enum would make every pattern unrepresentable. It is a named open string type — `SettingsTimeFormat` — with the presets as constants, and both the unit test and the integration test round-trip an actual pattern rather than only the presets.

**`managedMcpServers` keeps an untyped object value** (L5991). Upstream types the entry as `Record<string, unknown>`; narrowing it to a Go struct here would reject entry shapes the CLI accepts, and this is an admin-controlled surface where being wrong is expensive.

The remaining three are straightforward: `bashOutputMaxChars` (L6298, default 30000), `taskOutputMaxChars` (L6302, default 32000), both clamped 4000–128000 by the CLI, and `timeZone` (L8080).

## Comment-only change

`managedSourcesBehavior`'s opt-out list grew upstream. `sandbox.credentials.awsPairs` and `sandbox.ripgrep` join the settings owned whole by the highest source, and `managedMcpServers` unions by server name — but a name set by two sources takes the higher source's entry whole rather than merging the two. Updated the comment on the field #228 added.

## Testing

Probed the bare CLI with all six through `--managed-settings` before writing any live assertion. This matters: an unrecognized nested member stalls the handshake rather than erroring, which presents as a hang and costs the full timeout. It started clean.

Both integration tests **pass rather than skip** — unusual for a settings PR:

- `TestIntegrationSettingsOptions/v0_3_263_settings` and `/v0_3_263_time_format_pattern` assert the values survive the managed tier intact through real argv capture.
- `TestIntegrationSettingsV0_3_263RealCLI` starts the actual CLI on these settings and drives a turn to completion, which is the assertion that would have caught a handshake stall.

Verified locally: `TestIntegrationSettingsOptions` 10.58s, `TestIntegrationSettingsV0_3_263RealCLI` 3.50s, both PASS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ stdin/stdout, giving you access to Claude's tool use, extended thinking,
 session management, and hook system.
 
 This repository tracks the official TypeScript Agent SDK surface through the
-v0.3.241 catchup work, using Go idioms where the API shape differs.
+v0.3.263 catchup work, using Go idioms where the API shape differs.
 
 ```mermaid
 flowchart TB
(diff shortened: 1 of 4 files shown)


Roasbeef:
Self-review against `scratch/ts-sdk-0.3.263/package/sdk.d.ts`. Six fields, all verified at their cited lines: L5889 (nested in permissions), L5991, L6298, L6302, L8076, L8080.

Three calls worth defending:

**`SettingsTimeFormat` as an open string type, not an enum.** Upstream is `('auto' | '12-hour' | '24-hour' | '24-hour-utc') | string` — the presets are a convenience, not a closed set, and any value containing `%` is a strftime pattern. An enum would make every pattern unrepresentable, which is a silent capability loss rather than a compile error. Constants name the presets; the type accepts patterns. Round-tripped in both unit and integration tests so the pattern path is not theoretical.

**`ManagedMCPServers` keeps `map[string]map[string]interface{}`.** Upstream types the entry as `Record<string, unknown>`. Narrowing to a struct would reject entry shapes the CLI accepts, on an admin-controlled surface where a rejected config is an outage rather than an inconvenience.

**`BlockReadsOutsideWorkingDirectories` nests in the permissions block.** Filed at top level it would marshal to a key the CLI ignores, and a test that only asserted "the field survives" would still pass. The unit test asserts both directions — present under `permissions`, absent at top level.

Probed the bare CLI through `--managed-settings` with all six before writing a live assertion, since an unrecognized nested member stalls the handshake rather than erroring — that failure presents as a hang and costs the full timeout. It started clean, so both integration tests assert rather than skip: argv round-trip through the managed tier, and a real CLI start driving a turn to completion.

Also updated the `managedSourcesBehavior` comment, whose opt-out list grew upstream — `sandbox.credentials.awsPairs` and `sandbox.ripgrep` join the owned-whole group, and `managedMcpServers` unions by name with a contested name taken whole from the higher source.

Two process notes on this branch, both netted out by squash-merge: a `misspell` lint failure (CI enforces US spelling — "modelling"), and a README parity write-up that landed here by accident from a dirty tree and was reverted in a follow-up commit. Final diff against main is `options.go`, `options_test.go`, `integration_test.go` only.

VERDICT: APPROVE

PR #245, opened 2026-09-07
docs: refresh TypeScript SDK parity to v0.3.263

Roasbeef:
Parity documentation through v0.3.263.

This covers **two** cycles. The v0.3.251 docs refresh (PR-12 of that cycle) was dropped when the cycle closed out, so without this the README record jumps from v0.3.241 straight to v0.3.263 with a gap in the middle. Both sections land here.

Each entry leads with *why* the surface exists rather than restating the field list — the field list is already in the code and the doc comments. The parts worth reading twice:

- **Plugin delivery** is a Windows command-line limit (32,767 characters, one `--plugin-dir` per plugin). Notes that `argv` stays the Go default because `initialize` needs Claude Code 2.1.261+ and an older binary *exits at startup* on the unknown flag rather than degrading.
- **`user_message_uuids`** is documented as the gap `user_message_uuid` left open, not as a new field: the singular names only the last member of a merged batch. Also records that the result list can be strictly longer than the reply list.
- **`no_response`** is framed as a retry-budget reclassification rather than two durations.
- **`thinkingTokens`** carries both traps forward: contained in `outputTokens`, and partial across CLI versions so unsafe as a denominator.

Deferrals for both cycles are consolidated into one list rather than repeated per cycle, since all four carried across unchanged.

Also bumps the header line from v0.3.241 to v0.3.263. Swept `README.md` and `docs/` for stale version references; the only remaining v0.3.233/v0.3.241 mentions are historical section headers and the "deferred every cycle since" note, which are correct as written.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ stdin/stdout, giving you access to Claude's tool use, extended thinking,
 session management, and hook system.
 
 This repository tracks the official TypeScript Agent SDK surface through the
-v0.3.241 catchup work, using Go idioms where the API shape differs.
+v0.3.263 catchup work, using Go idioms where the API shape differs.
 
 ```mermaid
 flowchart TB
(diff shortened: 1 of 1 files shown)


Roasbeef:
Self-review. Documentation only — no code, no test changes.

Checked each claim in the v0.3.263 section against the merged PRs rather than against the plan, since the plan is what was intended and the PRs are what shipped. The v0.3.251 section is reconstructed from #218–#229 plus that cycle's recorded confirmed-facts, not from branch names.

Two things this deliberately does that a mechanical refresh would not:

**It covers two cycles.** The v0.3.251 docs refresh was dropped when that cycle closed out. Landing only v0.3.263 would leave the README jumping from v0.3.241 to v0.3.263 with the intervening release undocumented — and that gap gets harder to reconstruct every cycle it survives.

**Entries lead with why, not what.** The field list already exists in the code and the doc comments; a README that repeats it earns nothing. So plugin delivery is framed as a Windows command-line limit, `user_message_uuids` as the correlation gap `user_message_uuid` left open, and `no_response` as a retry-budget reclassification. The two `thinkingTokens` traps (contained in `outputTokens`, partial across CLI versions) are carried forward because they are the kind of thing a reader acts on.

Deferrals consolidated into one list rather than repeated per cycle — all four carried across both unchanged, and duplicating them would imply they were re-evaluated separately.

Swept `README.md` and `docs/` for stale version strings. Remaining v0.3.233 and v0.3.241 mentions are historical section headers and the "deferred every cycle since v0.3.233" note, correct as written.

VERDICT: APPROVE

Issue #246, opened 2026-09-14
Parity with TypeScript Agent SDK v0.3.270


Roasbeef:
Tracks the weekly catchup against `@anthropic-ai/claude-agent-sdk` v0.3.270 (npm, 2026-09-14). Prior shipped parity: v0.3.263 (#230).

The `sdk.d.ts` diff is 588 lines but only ~120 of them are non-comment; `sdk.mjs` emits an identical `--flag` set, and `agentSdkTypes.d.ts` / `bridge.d.ts` are byte-identical. So this is a types-and-wire cycle with zero argv work: two new control-request subtypes, a handful of message fields, settings parity, and one optional transport capability.

### Planned PRs

- [ ] PR-01 `messages:` `resume_reason` on assistant, stream and result frames
- [ ] PR-02 `messages:` `result_index` and `local_command` on result frames
- [ ] PR-03 `messages:` `verification_required` / `cloud_credential_error` assistant errors, `RateLimitInfo.limitScope`
- [ ] PR-04 `query:` `kind` on `get_context_usage` categories
- [ ] PR-05 `client:` `hold_on_cache_impact` option for `reload_plugins`
- [ ] PR-06 `client:` `list_permission_rules` control request
- [ ] PR-07 `client:` `get_hooks_listing` control request
- [ ] PR-08 `options:` settings parity (`bashEditDiffEnabled`, `prependPlugins`, `appendPlugins`, `gatewayInternalNetworks`, `maxEffortLevel`)
- [ ] PR-09 `protocol:` send `sdkMcpServerManifests` on initialize
- [ ] PR-10 `transport:` `DeliveryMarker` optional capability (`markDelivered`)
- [ ] PR-11 `docs:` refresh TypeScript SDK parity to v0.3.270

### Deferred

- `rename_session` `source` / `session_id` — the `rename_session` control request is not implemented in the Go SDK at all, so there is no field to add. Adding the RPC is its own surface, not a v0.3.270 divergence.
- `sdk-tools.d.ts` churn (104-line diff) — tool input/output schema drift, deferred every cycle since v0.3.215.
- Doc-comment-only rewords with no shape change (pending-request presence guarantees, meta-turn `user_message_uuid` fold semantics, `systemPromptSnapshot` default wording, `applyFlagSettings` session-reset keys). Folded into PR-11 where a Go doc comment already quotes the old text.

Roasbeef:
[...]

PR #247, opened 2026-09-14
messages: add resume_reason to assistant, stream and result frames

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-01".

When a worker restart interrupts a turn, the CLI re-runs it automatically under `CLAUDE_CODE_RESUME_INTERRUPTED_TURN`. Nothing on the wire distinguished that re-run's reply frames from the interrupted attempt's: both stamp `user_message_uuid`, and on a re-run that uuid names the *interrupted* turn's own last user prompt, so a consumer reconciling replies against its sends sees two frames claiming the same send.

- `ResumeReason` added to `AssistantMessage` (sdk.d.ts v0.3.270 L3362), `PartialAssistantMessage` (L5111) and `ResultMessage` (L5315 error, L5343 success). It rides the same frames as `user_message_uuid`; result frames carry it with or without an echo, since a re-run whose opener could not be vouched still has a reason to report.
- The value is the host's `CLAUDE_CODE_RESUME_REASON` when one was set, else the CLI's `interrupted_turn` fallback. That makes it an open set, so the field is a plain `string` and the four documented values get constants (`ResumeReasonInterruptedTurn`, `ResumeReasonHostDraining`, `ResumeReasonCheckpointRestore`, `ResumeReasonContainerRecreated`) — the same treatment `apiKeySource` gets.
- Not carried on `thinking_tokens` frames or on the Remote Control bridge's per-turn synthetic results; documented on the fields.

### Tests

- `TestParseMessageResumeReason` in `messages_test.go`: five subtests covering the assistant frame (with the interrupted turn's echo), the stream event, a success result with the `interrupted_turn` fallback and no echo, an error result, and an ordinary turn leaving the field empty.
- `TestIntegrationResumeReason` is a skip stub: reproducing the field needs a host that kills a worker mid-turn and brings it back on the same session, which the subprocess transport does not do. Recorded in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Self-review (orchestrator). Diff is 5 files, all additive: three struct fields, four constants, one unit test func, one integration skip stub.

Checked:
- Field placement matches the TS siblings — `resume_reason` sits next to `user_message_uuid(s)` on all three structs, which is where the stamping rule ties them together.
- Open-set treatment is right. The CLI forwards the host's `CLAUDE_CODE_RESUME_REASON` verbatim, so a closed Go type would reject valid values. Plain `string` + constants matches how `apiKeySource` is handled.
- The result-frame doc correctly records that the reason survives without an echoed uuid; that asymmetry is easy to miss and the test pins it.
- No parse-builder wiring needed — these structs unmarshal by tag, unlike `*HookInput`.

`go test ./...`, `go vet`, `gofmt -l`, `golangci-lint run` all clean locally; CI green on all 8 checks.

VERDICT: APPROVE

PR #248, opened 2026-09-14
messages: add result_index and local_command to result frames

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-02".

A stream-json consumer had no way to distinguish a lost result from a run that simply produced fewer of them.

- `ResultIndex` on `ResultMessage` (sdk.d.ts v0.3.270 L5320 error, L5382 success): how many results the run numbered before this one, from zero, in the order the process writes them. A result whose write fails still consumes its number, so a gap in the sequence is positive evidence a result was dropped. A result held back while background work finishes is numbered when it is finally written, not when its text was produced.
- Deliberately not `num_turns`: that counts model round-trips within one turn, this counts results across the run. The test asserts both on one frame so the distinction is pinned.
- Modeled as `*int` — zero is the first result, not a missing field, so a plain `int` would erase the difference between "first" and "older producer".
- `LocalCommand` on the same struct (L5344), success results only. `sdk.d.ts` declares it with no documentation and `sdk.mjs` never reads it, so it is carried through opaquely rather than guessed at; the doc comment says as much.

Numbering is done by the process hosting the run (`claude -p`, stream-json). A local client relaying a cloud session passes the cloud numbering through, and its own locally built error results carry none.

### Tests

- `TestParseMessageResultIndex` in `messages_test.go`: index zero surviving as a value, a gap alongside `num_turns`, an error result, `local_command`, and an older producer omitting both.
- `TestIntegrationResultIndex` is a skip stub — the runner CLI emits neither field, so a live assertion could only re-assert the absent state. Recorded in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Self-review (orchestrator). Additive: two fields on `ResultMessage`, one unit test func, one integration skip stub. Merged `origin/main` in to clear the EOF-append conflict with #247 (no force-push; both test blocks preserved and re-verified).

Checked:
- `*int` for `ResultIndex` is the right call and the test pins why: index zero is the first result, so a plain `int` would make "first" and "older producer omitted it" indistinguishable. Same reasoning `QueuedTurnCount` already uses on this struct.
- The `num_turns` contrast is real and easy to get wrong; asserting both on one frame in `a gap means a result was lost` documents it in code rather than only in a comment.
- `LocalCommand` is carried opaquely on purpose. sdk.d.ts declares it with no doc and sdk.mjs never reads it, so inventing semantics would be worse than passing it through. Doc comment says exactly that.

VERDICT: APPROVE

PR #249, opened 2026-09-14
messages: add assistant error codes and RateLimitInfo.limitScope

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-03".

- `AssistantMessageError` (sdk.d.ts v0.3.270 L3411) gains `verification_required` (turn refused pending account verification) and `cloud_credential_error` (the cloud provider credential could not be used). The latter trips gosec G101 on the string literal, so it carries a per-line `//nolint:gosec` — it is a wire enum value, not a credential.
- `RateLimitInfo.LimitScope` (L5270), typed as a new `RateLimitScope` with `service` / `channel` / `group_pool`.

`limitScope` matters more than its size suggests. A pooled group budget running out produces a denial byte-identical to the member exhausting their own monthly cap. Without the field, a host tells the user to buy credits when the actual remedy is to talk to whoever owns the team budget. The doc comment on the constant says so, since that is the whole reason the field exists.

Empty on a plain member denial and from older CLIs.

### Tests

- `TestRateLimitInfoLimitScope`: the `group_pool` denial alongside `credits_required`, the `service` and `channel` scopes, and a plain member denial leaving the field empty.
- `TestAssistantMessageErrorCodes`: both new codes parsing off an assistant frame.
- The existing `TestRateLimitEventMessageRoundTripAndParse` minimal subtest now also asserts `limitScope` is omitted when unset, keeping the `omitempty` pinned.
- Two skip stubs recorded in `INTEGRATION-FOLLOWUPS.md`: `limitScope` needs an exhausted service/channel/group-pool budget, which is not something a test run should create, and the two new error codes need a deterministic upstream failure the fixture cannot force.


Roasbeef:
Self-review (orchestrator). Merged `origin/main` in to clear the EOF-append conflict with #248; verified `git diff origin/main...HEAD` contains only this PR's own changes (5 files, 140 insertions).

Checked:
- `RateLimitScope` as its own named type rather than a bare string: the union is closed upstream (three members, no open-set language in the doc), so this is the case where a typed enum is correct — unlike `resume_reason` in #247, which is genuinely open.
- The `group_pool` doc comment earns its length. That denial is byte-identical to a personal-cap denial on every other field, so a host without this branches wrong and tells the user to buy credits. Worth stating on the constant, not just in the PR.
- `//nolint:gosec` is per-line, not block-level — block-level does not propagate for G101.
- The minimal-subtest key list now includes `limitScope`, which keeps the `omitempty` from silently regressing. That is the test that would actually catch a bad tag.

VERDICT: APPROVE

PR #250, opened 2026-09-14
query: add kind to get_context_usage categories

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-04".

`SDKControlGetContextUsageResponse.categories[]` (sdk.d.ts v0.3.270 L3675) gains a `kind` discriminator: `used` | `free` | `buffer` | `deferred`. Upstream documents it bluntly — classify on this, never on the English name — and the name is exactly what a consumer had to fall back on before, since `IsDeferred` only covers one of the four cases.

- The enum already exists. `ContextUsageCategoryKind` landed in v0.3.233 for `AssistantContextUsageCategory`, which is the same classification carried on a different wire shape; this reuses it rather than minting a parallel type.
- Required on the wire upstream, but tagged `omitempty` here: a CLI that predates the field leaves it empty, and that is worth distinguishing from a row the CLI deliberately classified.
- `ContextUsageCategory` picked up a doc comment while I was in there, since `Name` now needs the "do not classify on this" warning.

### Tests

- `TestContextUsageCategoryKind` in `client_introspection_test.go`: all four kinds off one response, the `deferred` row cross-checked against the older `IsDeferred` signal, and an older CLI omitting the field.
- `TestIntegrationContextUsageCategoryKind` is a real live test, not a stub — it primes a session, pulls a context report, and asserts every classified row falls in the known set and that `IsDeferred` never disagrees with `Kind`.

Ran it against the installed CLI: it self-skips with `CLI predates get_context_usage categories[].kind`, so the runner binary does not emit the field yet. The test activates on its own once the CLI catches up; no backfill needed, just a bump. Noted in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Self-review (orchestrator). Merged `origin/main` in to clear the EOF-append conflict with #249; `git diff origin/main...HEAD` shows only this PR's 4 files / 121 insertions.

Checked:
- Reusing `ContextUsageCategoryKind` rather than minting a second enum is right — v0.3.233 added it for `AssistantContextUsageCategory`, which is the same classification on a different wire shape. Two types for one concept would drift.
- `omitempty` despite the field being required upstream is a deliberate divergence, and the "older CLI omits kind" subtest is what documents it. The runner binary today is exactly that case, so this is not hypothetical.
- The `IsDeferred` / `Kind` cross-check appears in both the unit and integration test. That is the one overlap between the old and new signals, and it is where a wrong mapping would show up first.
- The integration test is real, not a stub: it primes a session, pulls a live report, and validates every classified row. It self-skips today because the runner CLI does not emit the field yet — recorded in INTEGRATION-FOLLOWUPS.md as a CLI bump rather than a backfill.

`go test ./...`, `go vet`, `gofmt -l`, `golangci-lint run` clean locally; CI Lint + Test green.

VERDICT: APPROVE

PR #251, opened 2026-09-14
client: add hold_on_cache_impact option to reload_plugins

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-05".

Applying a plugin reload can change the session's tool list, and when the conversation's prompt cache depends on that list the reload silently throws the cache away. The interactive `/reload-plugins` has always checked for this and asked for `--force`; the control request had no equivalent, so an SDK host could only discover the impact after paying for it.

- `SDKControlReloadPluginsRequest` gains `hold_on_cache_impact` (sdk.d.ts v0.3.270 L4541), surfaced as `ReloadPluginsOptions{HoldOnCacheImpact}`.
- `SDKControlReloadPluginsResponse` gains `Held` and `CacheImpact` (L4564): the plugin MCP servers applying would register or drop, and whether the LSP tool would come or go. `LSPToolChange` is a typed enum; its `may-add` / `may-remove` members mean the preview could not fully see the pending plugin set, so the change is possible rather than certain.
- `Stream.ReloadPlugins` grows a variadic options parameter, following the `GetContextUsageOptions` precedent, so existing no-arg callers keep compiling.

`Held` is `*bool` because three states matter and two of them are false-ish: the check ran and came back clean, versus a CLI that predates the option and applied the reload unchecked. Collapsing those would let a host report an unverified reload as a verified one. The `older CLI applied it unchecked` subtest pins that distinction.

Note the casing asymmetry upstream — camelCase `holdOnCacheImpact` on the TS method, snake_case `hold_on_cache_impact` on the wire. The wire spelling is what this sends.

### Tests

- `TestStreamReloadPluginsHoldOnCacheImpact` in `client_file_plugin_control_test.go`: the option on the request, its absence when no opts are passed, a held response with the impact preview parsed out, an explicit `held: false`, and an older CLI omitting the field entirely.
- `TestIntegrationReloadPluginsHoldOnCacheImpact` is a real live test. Ran it against the installed CLI: the request is **accepted without a protocol error** — the field is safe to send today — but the response omits `held`, so it self-skips. Assertions activate on a CLI bump. Holding on a genuine impact additionally needs a plugin fixture that changes the tool list; noted in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Self-review (orchestrator). Merged `origin/main` in to clear the EOF-append conflict with #250.

Checked:
- `Held *bool` is the load-bearing decision and the test suite pins all three states. Collapsing "check ran, came back clean" into "CLI ignored the option" would let a host report an unverified reload as verified — which is exactly the failure this feature exists to prevent.
- `LSPToolChange` as a typed enum with the `may-` variants preserved verbatim. Folding `may-add` into `adds` would overstate certainty the CLI explicitly declined to claim.
- Variadic options rather than a new method or a changed signature — matches `GetContextUsage`, and existing no-arg `ReloadPlugins(ctx)` callers keep compiling. The "no opts omits the option" subtest guards the wire side of that.
- `messages.go` shows 57 changed lines for a one-field addition; that is gofmt realigning the struct-tag column, verified by diffing the added lines. No semantic change beyond `HoldOnCacheImpact`.
- Live integration run confirms the CLI accepts the new field without a protocol error, so this is safe to send against today's binaries even though the check itself is not implemented yet. That is the useful result from the run, and it is in the PR body.

VERDICT: APPROVE

PR #252, opened 2026-09-14
client: add list_permission_rules control request

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-06".

`/permissions` has always been able to show a user which rules the session is actually running under and where each came from. An SDK host had no equivalent: it could *set* rules but never enumerate the effective set, so anything rendering a permissions surface had to reconstruct it from settings files and would miss session-only approvals, slash-command grants and `--allowedTools` rules entirely.

- New `Stream.ListPermissionRules` (sdk.d.ts v0.3.270 L4303) returning `SDKControlPermissionRulesState`.
- `PermissionRuleEntry` models provenance rather than flattening it: `Source` (11-member closed union, mirroring the CLI's `PermissionRuleSource` — upstream keeps a parity test against it), `Editability` (`persistent` / `session` / `readonly` — what changing the rule would take), and `NotInEffect` for rows a managed-settings lock has neutered. Those rows still appear, because a user needs to see that their rule exists *and* is being ignored.
- `PermissionRuleDescription` is split `prefix` / `emphasis` / `suffix` so a host can emphasize the rule-derived fragment the way the terminal does.
- `PermissionWorkspaceDirectory` for the `--add-dir` / `/add-dir` / IDE-workspace grants.
- `SDKSettingsParseError` lands here too. It is shared with `get_settings`, which this SDK does not implement yet, but the shape belongs to whichever surface needs it first.

Rule strings come back **verbatim** — invisible characters included — and two spellings that parse identically stay separate entries. Both are deliberate upstream, and the doc comments say to escape at display. The `rules are verbatim, not normalized` subtest pins both so a future cleanup does not quietly normalize them away.

`hostCredential` trips gosec G101; per-line `//nolint` as with the other wire enums.

### Tests

- `TestStreamListPermissionRules`: request wire shape, full parse across all three behaviors and editabilities, the verbatim/no-dedup guarantee, the managed-only case (`NotInEffect` + `errors` from a skipped settings file, cross-checked against the readonly invariant), and the error path.
- `TestIntegrationListPermissionRules` is a real live test — it launches the session with `--allowedTools Bash(git status:*)` and asserts that rule comes back with its provenance, rather than depending on whatever settings files the machine happens to have.

Ran it against the installed CLI: `Unsupported control request subtype: list_permission_rules`, so it self-skips. Activates on a CLI bump, no backfill needed. Noted in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Self-review (orchestrator).

Checked:
- `Source` and `Editability` as closed typed enums is correct here, unlike `resume_reason` in #247 — upstream keeps a parity test (`listPermissionRules.test.ts`) against the CLI's `PERMISSION_RULE_SOURCES`, which is an explicit stability commitment. Noted on the type.
- The verbatim/no-dedup guarantee is the part most likely to get "fixed" later, so it gets its own subtest with both an invisible-character rule and two identically-parsing spellings. The staticcheck ST1018 hit on the literal was resolved with a `\\u200b` escape, which keeps the test meaningful and the source greppable.
- `NotInEffect` rows being listed at all is the right call and worth the doc: a user whose rule is silently ignored by a managed-settings lock needs to see that it exists *and* is inert. The managed-only subtest asserts the readonly invariant that comes with it.
- `SDKSettingsParseError` defined here rather than deferred. It is shared with `get_settings` (not implemented in this SDK), but the shape has to live somewhere and this is its first consumer.
- Integration test asserts against a rule the session itself injects via `--allowedTools`, not against ambient machine settings — so it will be meaningful on any runner, not just this one.

`go test ./...`, `go vet`, `gofmt -l`, `golangci-lint run` clean locally; CI Lint + Test green.

VERDICT: APPROVE

PR #253, opened 2026-09-14
client: add get_hooks_listing control request

Roasbeef:
Part of #246 (TypeScript SDK v0.3.270 parity). See `memory/catchup-v0.3.270/PLAN.md` §"PR-07".

Hooks have been write-only from the SDK's side: a host could register them but never ask what the session was actually running. Reconstructing the answer from settings files gets it wrong in every interesting case — plugin hooks, session hooks, the managed-policy locks, and files the merge skipped for a parse error.

New `Stream.GetHooksListing` (sdk.d.ts v0.3.270 L3766) returning the `/hooks` menu's own data: `Events` (events with at least one hook, lifecycle order), `Hooks` (one row per hook, with an optional `Editable` target for `claude edit-hook`), `EventCatalog` (every event, for an add-hook form), `Policy`, and the optional `SafeMode` / `BareMode` banners plus skipped-settings `Errors`.

Two things in the response are easy to misread, and both carry doc comments saying so:

- **An empty `Hooks` list under `Policy.ManagedOnly` does not mean nothing fires.** Managed hooks are deliberately withheld from the listing while still running; `PolicyHookCount` is how you tell. The `managed-only withholds hooks that still run` subtest asserts exactly that combination.
- **`Editable.HeadersRedacted` marks headers the CLI blanked, not headers the user cleared.** A replace that sends no headers keeps the stored ones, so a host that treats the blanks as intent would silently wipe them.

`Disabled` and `HeadersRedacted` are `true`-only optionals on the wire, so `bool` + `omitempty` is exact here rather than lossy — absent and `false` genuinely mean the same thing.

Strings arrive display-ready (the producer has already revealed control characters), which the `HookListingEntry` doc records so a host does not double-escape. They are still author-controlled — a plugin picks its own command text.

### Tests

- `TestStreamGetHooksListing`: request wire shape; a full parse across a command hook, a plugin HTTP hook with redacted headers, and a disabled prompt hook with a condition and no edit target; the managed-only case; the safe-mode/bare-mode banners with a whole-file parse error; and the error path.
- `TestIntegrationGetHooksListing` is a real live test. It asserts against the **event catalog** rather than hook rows — the catalog is the binary's own event list, so it is non-empty on any session regardless of machine configuration — plus the cross-references that every listed event is catalogued and every hook row names a listed event.

Ran it against the installed CLI: `Unsupported control request subtype: get_hooks_listing`, so it self-skips. Activates on a CLI bump. Noted in `INTEGRATION-FOLLOWUPS.md`.


Roasbeef:
Superseded by #255, which merged the same `get_hooks_listing` surface 12 minutes after this PR was opened. Both came out of the v0.3.270 cycle when two orchestrator instances raced on the shared tree — the same collision that produced the duplicate tracking issue #246 alongside #254.

#255 is the version on `main`. Closing this one as the duplicate; no work is lost.

Issue #254, opened 2026-09-14
Parity with TypeScript Agent SDK v0.3.270


Roasbeef:
Weekly TS SDK catchup: `@anthropic-ai/claude-agent-sdk` is at **0.3.270**; Go SDK last shipped parity at **0.3.263** (#245).

Most of the 0.3.264–0.3.270 delta already landed pre-cycle: #247–#252. Remaining surfaces:

- [ ] PR-01 — `get_hooks_listing` control request + response (`SDKControlGetHooksListing{Request,Response}`)
- [ ] PR-02 — `sdkMcpServerManifests` field on the `initialize` control request
- [ ] PR-03 — Settings parity v0.3.270: `bashEditDiffEnabled`, `prependPlugins`, `appendPlugins`, `gatewayInternalNetworks`, `maxEffortLevel` (top-level + per-model `SettingsModel`)
- [ ] PR-04 — docs parity refresh to v0.3.270 (behavioral doc deltas: systemPromptSnapshot default flip, applyFlagSettings reset keys, user_message_uuid fold semantics on assistant/partial/error-result/thinking frames, usage weekly-window absent-vs-empty, pending_* v2.1.268 presence, parentSettingsBehavior merge expansion, file-tool deny-rule note)

Deferred (standing gaps, not this cycle):
- `rename_session` `source`/`session_id` — the Go SDK has no `rename_session` control request.
- TS `Query.markDelivered` transport hook — no Go transport analog.

Roasbeef:
v0.3.270 parity complete. Merged this cycle:

- #255 — `get_hooks_listing` control request + response
- #257 — Settings parity (`bashEditDiffEnabled`, `prependPlugins`, `appendPlugins`, `gatewayInternalNetworks`, `maxEffortLevel` top-level + per-model)
- #258 — docs parity refresh to v0.3.270

Landed pre-cycle (0.3.264–0.3.270 tail): #247–#252.

Deferred:
- #256 — `sdkMcpServerManifests` on initialize (no in-process handshake to cache; additive + degrades gracefully).
- `rename_session` `source`/`session_id` and TS `Query.markDelivered` — no Go analogue.

`main` at 136b48d.

PR #255, opened 2026-09-14
client: add get_hooks_listing control request

Roasbeef:
Part of #254 (TS SDK v0.3.270 parity). New read-only control request `get_hooks_listing` (`SDKControlGetHooksListingRequest` / `SDKControlGetHooksListingResponse`, sdk.d.ts v0.3.270 L3672–3907).

## What
- `Stream.GetHooksListing(ctx)` — mirrors `Stream.ListPermissionRules` (#252): sends `{subtype: "get_hooks_listing"}`, returns the parsed response.
- Response modeled with named structs (`HooksListingEvent`, `HooksListingHook`, `HooksListingEditable`, `HooksListingEventCatalogEntry`, `HooksListingPolicy`, `HooksListingSafeMode`, `HooksListingBareMode`) so the optional presence flags (`disabled`, `headersRedacted`, `safeMode`, `bareMode`) stay typed rather than collapsing into `map[string]any`.
- Reuses `SDKSettingsParseError` (added in #252) for skipped-settings-file errors.

## Why
Parity gap: a host could set hooks via settings but had no way to enumerate the effective, merged set the `/hooks` menu renders — settings-file + session + plugin hooks, plus the policy/safe-mode state that decides which run. This is the hooks analogue of `list_permission_rules`.

## Notes
- Display strings arrive with control characters revealed (not escaped) — doc comments direct callers to escape at their own surface, same as permission-rule strings.
- `editable` is the round-trip target for a host edit form / `claude edit-hook`; HTTP header values come back blanked (`headersRedacted`).

## Tests
- `client_introspection_test.go`: request wire shape; full nested round-trip (events, hooks, editable with/without `headersRedacted`, eventCatalog, policy, safeMode, errors); error path.
- `integration_test.go`: `TestIntegrationGetHooksListing` asserts the event catalog is non-empty and per-event `hookCount` agrees with the listed rows. Skips on a CLI predating the subtype (installed CLI 2.1.222 skips).


Roasbeef:

- Additive-only: new `Stream.GetHooksListing` + response structs, no change to existing surfaces.
- Wire shape matches sdk.d.ts v0.3.270 L3672–3907; nested optional presence flags (`disabled`, `headersRedacted`, `safeMode`/`bareMode`) modeled as typed structs/bools, not `map[string]any`.
- Reuses `SDKSettingsParseError` from #252 rather than redefining it.
- Client method is a faithful mirror of `ListPermissionRules` (same marshal/unmarshal/error pattern).
- Tests cover request wire shape, full nested round-trip, and error path; integration test asserts eventCatalog invariant and per-event hookCount vs. rows, skips on an older CLI (installed 2.1.222 skips).
- gofmt/vet/`go test ./...` green; CI green across all Build matrix entries + Lint + Test.

VERDICT: APPROVE

PR #257, opened 2026-09-14
options: settings parity with TypeScript SDK v0.3.270

Roasbeef:
Part of #254 (TS SDK v0.3.270 parity). Six `Settings` field additions.

## Fields (sdk.d.ts v0.3.270)
- `bashEditDiffEnabled *bool` (L6682) — Bash-tool changed-file diff; PostToolUse Bash hooks get the changed-file list in `tool_response`. On by default where the Bash tool handles edits, so a pointer is what expresses "off".
- `prependPlugins []string` (L6789) / `appendPlugins []string` (L6793) — managed-plugin hook ordering: prepend runs first/outermost, append runs last/innermost. Honored only from managed (or own-machine user) settings.
- `gatewayInternalNetworks []string` (L7992) — the IPv4 CIDR blocks a Cloud gateway sits in; the `/login` internal-network gate. Admin-managed only.
- `maxEffortLevel EffortLevel` top-level (L8329) and per-model `SettingsModel.MaxEffortLevel` (L8342) — the effort ceiling. Unlike the applied `EffortLevel`, `"max"` **is** a member here: it's how you decline to cap (top-level) or exempt one model (per-model).

## Notes
- Completes and lands WIP that was sitting uncommitted in the working tree (all six fields, cites verified exact against sdk.d.ts v0.3.270); this PR adds the tests it was missing and the RealCLI/followups coverage.
- All top-level keys — an older CLI ignores them cleanly rather than stalling the managed-settings handshake (which is how an unknown *nested source-union* member fails).

## Tests
- `options_test.go`: `TestSettingsParityV0_3_270` (round-trip all six, including `"max"` on the per-model cap) + `TestSettingsParityV0_3_270OmitEmpty`.
- `integration_test.go`: `TestIntegrationSettingsV0_3_270RealCLI` — verified live: the CLI starts and completes a turn on these settings (12s, no fault). Managed-only observation of the ordering/gateway fields deferred to an MDM fixture (INTEGRATION-FOLLOWUPS.md).


Roasbeef:

- Six additive Settings fields; all cites verified exact against sdk.d.ts v0.3.270 (L6682/L6789/L6793/L7992/L8329/L8342).
- `bashEditDiffEnabled` correctly a `*bool` (default-on field needs the pointer to express "off"); the rest are `[]string` / `EffortLevel` with omitempty.
- Per-model `MaxEffortLevel` semantics right: `"max"` is a member (exempts a model), distinct from the applied `EffortLevel` — doc comment and test both pin it.
- Unit tests cover round-trip + omitempty for all six; RealCLI integration test verified live (CLI starts + completes a turn, no fault); managed-only observation gap tracked in INTEGRATION-FOLLOWUPS.
- gofmt/vet/`go test ./...` green; CI green across Build matrix + Lint + Test.

VERDICT: APPROVE

PR #258, opened 2026-09-14
docs: refresh TypeScript SDK parity to v0.3.270

Roasbeef:
Part of #254 (TS SDK v0.3.270 parity). Comment-only — no wire or behavior change; `git diff -U0` is a pure comment diff.

Mirrors the behavioral doc deltas 0.3.263 → 0.3.270 onto the Go fields that already carry them:

- **`SystemPromptSnapshot`** (options.go): recording is now the default — `nil` behaves like `true`, not "off". The old three-state description was actively misleading post-v0.3.270. `WithSystemPromptSnapshot` updated to match.
- **`ApplyFlagSettings`** (client.go): `effortLevel`/`model`/`agent`/`ultracode` clear to a *session reset* rather than a lower-precedence value; `effortLevel: "max"` is session-only and clamped to the org effort limit (sdk.d.ts L2674).
- **`UserMessageUUID` / `UserMessageUUIDs`** on assistant + result frames (messages.go): meta-turn fold semantics — a folded user message takes the echo over; a meta turn's own uuid is echoed only when the host vouches it (sdk.d.ts L3352/L3356/L5305).
- **`UsageRateLimits.ModelScoped`** (query_types.go): `nil` (unknown/cached/allowlist-hidden) is distinct from a present-but-empty slice (endpoint answered, listed none) (sdk.d.ts L4010).
- **Pending permission/dialog requests** (messages.go): always present (possibly empty) on a success initialize from CLI v2.1.268+; `nil` means an older CLI, and an inherited prompt can stay answerable without appearing here (sdk.d.ts L306–L333).

Cites bumped to v0.3.270 line numbers. `gofmt`/`go vet`/`go test ./...` green.


Roasbeef:

- Pure comment diff (`git diff -U0` shows no code lines); every edit mirrors a v0.3.270 behavioral doc delta onto a field that already carries the prose.
- Highest-value fix is `SystemPromptSnapshot`: the pre-v0.3.270 "omitted = off" three-state description was actively misleading once recording became the default — now `nil`≡`true`.
- Cites bumped to v0.3.270 line numbers; spot-checked against sdk.d.ts.
- Follow-up commit fixes a `misspell` lint (`signalled`→`signaled`); Lint + Test + Build matrix all green.

VERDICT: APPROVE

